PR #1, opened 2023-01-10
Remove unneeded files and add them to `.gitignore`

katzuv:
The files which were removed from the `.idea` directory are auto-generated and should not be tracked. They are also all of the files in this directory which are tracked, so the whole directory is being deleted.
Adding them to `.gitignore`, actively un-tracking them is needed too. You can read more [here](https://stackoverflow.com/a/1274447/10015776).


GaiaZano05 (approved):

katzuv:
Fixed conflict on file `.idea/jarRepositories.xml` so now this PR can be merged.
The conflict arose because this file was edited on `main` but deleted on this branch. By telling Git this file should really be deleted (because it shouldn't be version-controlled) we fixed the conflict.

katzuv:
The CI failed because of syntax errors in `PrototypeArm.java`. This is irrelevant to this branch so merging anyways.
@GaiaZano05, thanks for the approval 😄.

PR #3, opened 2023-01-11
Create vision subsystem

Emma03L:
Use limelight to estimate position on the field and find distance, angle and apriltag id from target.
Use LoggedSubsystem to log limelight

### DIFF
--- a/src/main/java/frc/robot/subsystems/vision/Constants.java
+++ b/src/main/java/frc/robot/subsystems/vision/Constants.java
@@ -6,5 +6,7 @@ import edu.wpi.first.math.geometry.Rotation3d;
 public class Constants {
     public static final double CAMERA_HEIGHT = 1.1; //[m]
     public static final double CAMERA_PITCH = -0.611; //[radian]
-    public static final Pose3d UPPER_CONE_LEFT_GRID_TARGET = new Pose3d(16.178, 4.99, 1.17, new Rotation3d(0, 0,0)); //[m]
+    public static final double UPPER_CONE_TARGET_TAPE_HEIGHT = 1.17; //[m]
+    public static final double LOWER_CONE_TARGET_TAPE_HEIGHT = 0.65; //[m]
+    public static final Pose3d UPPER_CONE_LEFT_GRID_TARGET = new Pose3d(16.178, 4.99, 1.17, new Rotation3d());
 }
--- a/src/main/java/frc/robot/subsystems/vision/Constants.java
+++ b/src/main/java/frc/robot/subsystems/vision/Constants.java
@@ -1,9 +1,11 @@
 package frc.robot.subsystems.vision;
 
+import edu.wpi.first.math.geometry.Pose3d;
+import edu.wpi.first.math.geometry.Rotation3d;
+
 public class Constants {
     public static final double CAMERA_HEIGHT = 1.1; //[m]
     public static final double CAMERA_PITCH = -0.611; //[radian]
     public static final double TARGET_HEIGHT = 0.65; //[m]
-    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0; //[m]
-    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0; //[m]
+    public static final Pose3d UPPER_CONE_LEFT_GRID_TARGET = new Pose3d(16.178, 4.99, 1.17, new Rotation3d(0, 0,0)); //[m]
 }
--- a/src/main/java/frc/robot/subsystems/vision/Constants.java
+++ b/src/main/java/frc/robot/subsystems/vision/Constants.java
@@ -1,7 +1,7 @@
 package frc.robot.subsystems.vision;
 
-public class VisionConstants {
-    public static final double CAMERA_HEAIGHT = 0;
+public class Constants {
+    public static final double CAMERA_HEIGHT = 0;
     public static final double CAMERA_PITCH = 0;
     public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0;
     public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0;
--- a/src/main/java/frc/robot/subsystems/vision/Constants.java
+++ b/src/main/java/frc/robot/subsystems/vision/Constants.java
@@ -6,6 +6,5 @@ import edu.wpi.first.math.geometry.Rotation3d;
 public class Constants {
     public static final double CAMERA_HEIGHT = 1.1; //[m]
     public static final double CAMERA_PITCH = -0.611; //[radian]
-    public static final double TARGET_HEIGHT = 0.65; //[m]
     public static final Pose3d UPPER_CONE_LEFT_GRID_TARGET = new Pose3d(16.178, 4.99, 1.17, new Rotation3d(0, 0,0)); //[m]
 }
--- a/src/main/java/frc/robot/subsystems/vision/Constants.java
+++ b/src/main/java/frc/robot/subsystems/vision/Constants.java
@@ -1,8 +1,9 @@
 package frc.robot.subsystems.vision;
 
 public class Constants {
-    public static final double CAMERA_HEIGHT = 0;
-    public static final double CAMERA_PITCH = 0;
-    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0;
-    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0;
+    public static final double CAMERA_HEIGHT = 1.1; //[m]
+    public static final double CAMERA_PITCH = -0.611; //[radian]
+    public static final double TARGET_HEIGHT = 0.65; //[m]
+    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0; //[m]
+    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0; //[m]
 }
--- a/src/main/java/frc/robot/subsystems/vision/Limelight.java
+++ b/src/main/java/frc/robot/subsystems/vision/Limelight.java
@@ -1,6 +1,7 @@
 package frc.robot.subsystems.vision;
 
 import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Pose3d;
 import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.geometry.Translation2d;
 import edu.wpi.first.net.PortForwarder;
@@ -22,17 +23,14 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
 
     private Limelight() {
         super(new LimelightLogInputs());
-        PortForwarder.add(5800, "limelight.local", 5800);
-        PortForwarder.add(5801, "limelight.local", 5801);
-        PortForwarder.add(5802, "limelight.local", 5802);
-        PortForwarder.add(5803, "limelight.local", 5803);
-        PortForwarder.add(5804, "limelight.local", 5804);
-        PortForwarder.add(5805, "limelight.local", 5805);
+        for (int i=5800; i<=5805; i++){
+            PortForwarder.add(i, "limelight.local", i);
+        }
     }
 
     /**
-     * if there is no instance of the limelight class it creates one and returns it
-     * @return
+     * If there is no instance of the limelight class it creates one and returns it
+     * @return Limelight instance
      */
     public static Limelight getInstance() {
         if (INSTANCE == null) {
@@ -42,16 +40,15 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
     }
 
     /**
-     * checks weather the limelight can see any targets
-     * @return
+     * Checks weather the limelight can see any targets
+     * @return weather the limelight detects any targets
      */
     public boolean hasTargets() {
         return tv.get() == 1;
     }
 
     /**
-     * returns the id of the aprilTag the limelight sees
-     * @return
+     * @return aprilTag id
      */
     public long getTagId() {
         return tid.get();
@@ -59,16 +56,14 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
 
 
     /**
-     * returns the pitch from the robot to the target
-     * @return
+     * @return pitch from robot to target
      */
     public Rotation2d getPitch() {
         return Rotation2d.fromDegrees(ty.get());
     }
 
     /**
-     *returns the distance from the target
-     * @return
+     * @return distance from target
      */
     public OptionalDouble getTargetDistance() {
         double totalPitch = Constants.CAMERA_PITCH + getPitch().getRadians();
@@ -80,8 +75,7 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
     }
 
     /**
-     * returns the yaw
-     * @return
+     * @return robot yaw
      */
     public OptionalDouble getYaw() {
         if (hasTargets()) {
@@ -91,30 +85,30 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
     }
 
     /**
-     * estimates the position of the robot
-     * @param robotAngle
-     * @return
+     * Estimates the position of the robot
+     * @param robotAngle The angle of the robot
+     * @return Optional Pose2d of the robot coordinates
      */
     public Optional<Pose2d> estimatePose(Rotation2d robotAngle) {
         double absoluteAngle = robotAngle.getRadians() + getYaw().orElse(0);
         double xTargetDistance = getTargetDistance().orElse(0) * Math.sin(absoluteAngle);
         double yTargetDistance = getTargetDistance().getAsDouble() * Math.cos(absoluteAngle);
-        if (hasTargets()) {
-            double xDistance = xTargetDistance - Constants.UPPER_CONE_TARGET11_X_DISTANCE;
-            double yDistance = yTargetDistance - Constants.UPPER_CONE_TARGET11_Y_DISTANCE;
-            Translation2d translation2d = new Translation2d(xDistance, yDistance);
-            return Optional.of(new Pose2d(translation2d, robotAngle));
+        if (!hasTargets()) {
+            return Optional.empty();
         }
-        return Optional.empty();
+        double xDistance = xTargetDistance - Constants.UPPER_CONE_LEFT_GRID_TARGET.getX();
+        double yDistance = yTargetDistance - Constants.UPPER_CONE_LEFT_GRID_TARGET.getY();
+        Translation2d translation2d = new Translation2d(xDistance, yDistance);
+        return Optional.of(new Pose2d(translation2d, robotAngle));
     }
 
     @Override
     public String getSubsystemName() {
-        return "Vision";
+        return "Limelight";
     }
 
     /**
-     * updates the limelight log inputs
+     * Updates the limelight log inputs
      */
     @Override
     public void updateInputs() {
--- a/src/main/java/frc/robot/subsystems/vision/Limelight.java
+++ b/src/main/java/frc/robot/subsystems/vision/Limelight.java
@@ -1,0 +1,69 @@
+package frc.robot.subsystems.vision;
+
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.net.PortForwarder;
+import edu.wpi.first.networktables.NetworkTable;
+import edu.wpi.first.networktables.NetworkTableEntry;
+import edu.wpi.first.networktables.NetworkTableInstance;
+import frc.robot.Constants;
+import frc.robot.subsystems.LoggedSubsystem;
+
+public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
+    public static Limelight INSTANCE = null;
+
+    private final NetworkTable table = NetworkTableInstance.getDefault().getTable("limelight");
+    private final NetworkTableEntry tx = table.getEntry("tx");
+    private final NetworkTableEntry ty = table.getEntry("ty");
+    private final NetworkTableEntry tv = table.getEntry("tv");
+    private final NetworkTableEntry ts = table.getEntry("ts");
+    private final NetworkTableEntry tid = table.getEntry("tid");
+    private LimelightLogInputs inputs = LimelightLogInputs.getInstance();
+
+    public Limelight getINSTANCE(){
+        if (INSTANCE==null){
+            INSTANCE = new Limelight();
+        }
+        return INSTANCE;
+    }
+
+    private Limelight() {
+        super(new LimelightLogInputs());
+        inputs = LimelightLogInputs.getInstance();
+        PortForwarder.add(5801, "limelight.local", 5801);
+    }
+
+    public boolean hasTargets(){
+        return tv.getDouble(0)!=0;
+    }
+
+    public int getTagId(){
+        return tid.getHandle();
+    }
+
+    public Rotation2d getPitch(){
+        return new Rotation2d(ty.getDouble(0));
+    }
+
+    public double getTargetDistance(){
+        double totalPitch = VisionConstants.CAMERA_PITCH+getPitch().getRadians();
+        return tx.getDouble(0)*Math.sin(totalPitch);
+    }
+
+    public Rotation2d getYaw(){
+        return new Rotation2d(ts.getDouble(0));
+    }
+
+    @Override
+    public String getSubsystemName() {
+        return "Limelight";
+    }
+
+    @Override
+    public void updateInputs() {
+        inputs.hasTargets = hasTargets();
+        inputs.yaw = getYaw().getRadians();
+        inputs.tagId = getTagId();
+        inputs.targetDistance = getTargetDistance();
+
+    }
+}
--- a/src/main/java/frc/robot/subsystems/vision/Limelight.java
+++ b/src/main/java/frc/robot/subsystems/vision/Limelight.java
@@ -25,27 +25,46 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
         PortForwarder.add(5801, "limelight.local", 5801);
     }
 
+    /**
+     * if there is no instace of the limelight class it creates one and returns it
+     * @return
+     */
     public Limelight getInstance() {
-
-
         if (INSTANCE == null) {
             INSTANCE = new Limelight();
         }
         return INSTANCE;
     }
 
+    /**
+     * checks weather the limelight can see any targets
+     * @return
+     */
     public boolean hasTargets() {
         return tv.get();
     }
 
+    /**
+     * returns the id of the aprilTag the limelight sees
+     * @return
+     */
     public long getTagId() {
         return tid.get();
     }
 
+
+    /**
+     * returns the pitch from the robot to the target
+     * @return
+     */
     public Rotation2d getPitch() {
         return new Rotation2d(ty.get());
     }
 
+    /**
+     *returns the distance from the target
+     * @return
+     */
     public OptionalDouble getTargetDistance() {
         double totalPitch = Constants.CAMERA_PITCH + getPitch().getRadians();
         if (hasTargets()) {
@@ -54,6 +73,10 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
         return OptionalDouble.empty();
     }
 
+    /**
+     * returns the yaw
+     * @return
+     */
     public OptionalDouble getYaw() {
         if (hasTargets()) {
             return OptionalDouble.of(ts.get());
@@ -61,6 +84,11 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
         return OptionalDouble.empty();
     }
 
+    /**
+     * estimates the position of the robot
+     * @param robotAngle
+     * @return
+     */
     public Optional<Pose2d> estimatePose(Rotation2d robotAngle) {
         double absouluteAngle = robotAngle.getRadians() + getYaw().getAsDouble();
         double xTargetDistance = getTargetDistance().getAsDouble() * Math.sin(absouluteAngle);
@@ -74,6 +102,9 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
         return Optional.empty();
     }
 
+    /**
+     * updates the limelight log inputs periodically
+     */
     public void periodic() {
         updateInputs();
     }
@@ -83,6 +114,9 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
         return "Limelight";
     }
 
+    /**
+     * updates the limelight log inputs
+     */
     @Override
     public void updateInputs() {
         loggerInputs.hasTargets = hasTargets();
--- a/src/main/java/frc/robot/subsystems/vision/Limelight.java
+++ b/src/main/java/frc/robot/subsystems/vision/Limelight.java
@@ -22,7 +22,12 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
 
     private Limelight() {
         super(new LimelightLogInputs());
+        PortForwarder.add(5800, "limelight.local", 5800);
         PortForwarder.add(5801, "limelight.local", 5801);
+        PortForwarder.add(5802, "limelight.local", 5802);
+        PortForwarder.add(5803, "limelight.local", 5803);
+        PortForwarder.add(5804, "limelight.local", 5804);
+        PortForwarder.add(5805, "limelight.local", 5805);
     }
 
     /**
--- a/src/main/java/frc/robot/subsystems/vision/Limelight.java
+++ b/src/main/java/frc/robot/subsystems/vision/Limelight.java
@@ -26,7 +26,7 @@ public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
     }
 
     /**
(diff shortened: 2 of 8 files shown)


src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,92 @@
...
+import edu.wpi.first.net.PortForwarder;
+import edu.wpi.first.networktables.NetworkTable;
+import edu.wpi.first.networktables.NetworkTableEntry;
+import edu.wpi.first.networktables.NetworkTableInstance;
+import frc.robot.subsystems.LoggedSubsystem;
+
+import java.util.Optional;
+import java.util.OptionalDouble;
+
+public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
```
katzuv, line 14: Could you please explain why is `LoggedSubsystem` required? I don't see 6328 needing it so I'd love to understand.
I suggest you take a look at [their vision code implementation from last year](https://github.com/Mechanical-Advantage/RobotCode2022/tree/main/src/main/java/frc/robot/subsystems/vision).

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,92 @@
...
+    private final NetworkTableEntry tid = table.getEntry("tid");
+    private LimelightLogInputs inputs = LimelightLogInputs.getInstance();
+
+    private Limelight() {
+        super(new LimelightLogInputs());
+        inputs = LimelightLogInputs.getInstance();
+        PortForwarder.add(5801, "limelight.local", 5801);
+    }
+
+    public Limelight getINSTANCE() {
```
katzuv: Method names in Java are written in `camelCase`.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,92 @@
...
+    private final NetworkTableEntry ty = table.getEntry("ty");
+    private final NetworkTableEntry tv = table.getEntry("tv");
+    private final NetworkTableEntry ts = table.getEntry("ts");
+    private final NetworkTableEntry tid = table.getEntry("tid");
+    private LimelightLogInputs inputs = LimelightLogInputs.getInstance();
+
+    private Limelight() {
+        super(new LimelightLogInputs());
+        inputs = LimelightLogInputs.getInstance();
+        PortForwarder.add(5801, "limelight.local", 5801);
```
katzuv: Do you understand what this line does and why it is required?

_(And on a personal note – please don't take this comment and others I'll write in the season as condescending. My only intent is that you'll learn and it's important to me to see you fully understand the code you're writing ☺️.
[IIRC](https://www.dictionary.com/browse/iirc) you haven't learned networking basics, so I'm asking the question above.)_

src/main/java/frc/robot/subsystems/vision/VisionConstants.java
```diff
@@ -0,0 +1,8 @@
+package frc.robot.subsystems.vision;
+
+public class VisionConstants {
```
katzuv: This class is inside the `vision` package, so calling it `Constants` is enough. Don't forget the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle.

src/main/java/frc/robot/subsystems/vision/VisionConstants.java
```diff
@@ -0,0 +1,8 @@
...
+    public static final double CAMERA_HEAIGHT = 0;
```
katzuv: Fix "heaight" spelling.

src/main/java/frc/robot/subsystems/vision/VisionConstants.java
```diff
@@ -0,0 +1,8 @@
...
+    public static final double CAMERA_PITCH = 0;
+    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0;
+    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0;
```
katzuv: 1. What does `11` mean? Is it the specific cone node?
2. What does "distance" mean? Distance from what?
3. [IIUC](https://en.wiktionary.org/wiki/IIUC), this is the location of the retroreflective tape on the cone node. If so, you should use WPILib's [`Pose3d`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/geometry/Pose3d.html).

src/main/java/frc/robot/subsystems/vision/LimelightLogInputs.java
```diff
@@ -0,0 +1,36 @@
+package frc.robot.subsystems.vision;
+
+import org.littletonrobotics.junction.LogTable;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+public class LimelightLogInputs implements LoggableInputs {
```
katzuv, line 15: You should use the `@AutoLog` annotation which was introduced in the 2023 version of AdvantageKit. Read more [here](https://github.com/Mechanical-Advantage/AdvantageKit/blob/f80d5531d7e44618f622c4079bb38032053b13a4/docs/NEW-FOR-2023.md#autolog-annotation).

katzuv (changes requested):
Good job! 😄 A few notes:
- Because f574cb2 comes right after 9a15126, you could actually merge them into one commit using `git commit --amend`. Read more about it [here](atlassian.com/git/tutorials/rewriting-history#git-commit--amend). This can be done also in GitKraken or in IntelliJ IDEA.
- The branch name is good 💯, just needs renaming from `feture` to `feature`; you can see [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch) how this can be done in GitHub and how to update the branch in your local repository. Also, otice that because you didn't provide any title for the PR, GitHub used the branch name, which isn't a good name. Please edit the PR's title to something more descriptive.

I'm available for any questions. Good luck! 💪🏻

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,92 @@
...
+
+public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
+    public static Limelight INSTANCE = null;
+
+    private final NetworkTable table = NetworkTableInstance.getDefault().getTable("limelight");
+    private final NetworkTableEntry tx = table.getEntry("tx");
+    private final NetworkTableEntry ty = table.getEntry("ty");
+    private final NetworkTableEntry tv = table.getEntry("tv");
+    private final NetworkTableEntry ts = table.getEntry("ts");
+    private final NetworkTableEntry tid = table.getEntry("tid");
```
katzuv: The NetworkTables protocol was upgraded to version 4 this year. [Here](https://docs.wpilib.org/en/stable/docs/software/networktables/nt4-migration-guide.html) is a guide on how to migrate to the new version.

src/main/java/frc/robot/subsystems/vision/Limelight.java
Emma03L: Gaya told me to put it in the constructor, I know it's used to forward the connection from the raspberry Pi that is used in the limelight.

src/main/java/frc/robot/subsystems/vision/VisionConstants.java
Emma03L: The names are only temporary.
These constants are supposed to be the x, y coordinates of a specific cone node.
11 means the left node in the leftmost grid section.

src/main/java/frc/robot/subsystems/vision/Limelight.java
Emma03L: I don't think it's required but Eitan said we should use it

src/main/java/frc/robot/subsystems/vision/Limelight.java
katzuv: Why did you resolve this thread?

src/main/java/frc/robot/subsystems/vision/Limelight.java
katzuv: This is not a good explanation.
Even if Dean Kamen came and told you to use `LoggedSubsystem`, I want you to know exactly why it should be used.

src/main/java/frc/robot/subsystems/vision/Limelight.java
katzuv: It's more accurate to say "forward the port", not the connection (that's also how it's written in the code).
Why is this important?

src/main/java/frc/robot/subsystems/vision/VisionConstants.java
katzuv: > The names are only temporary.

They still need to be clear.
> These constants are supposed to be the x, y coordinates of a specific cone node.

What about the Z axis?
> 11 means the left node in the leftmost grid section.

This is not clear from the name. You should add documentation that explains the naming.

src/main/java/frc/robot/subsystems/vision/LimelightLogInputs.java
katzuv: You didn't fully utilize the annotation. Please read the documentation again. I'm here to help ☺️.
Also, please don't resolve threads. It's the _reviewer_, which opened the thread, that closed threads, after confirming the problem was fixed.

src/main/java/frc/robot/subsystems/vision/Limelight.java
barel13: During competitions, as the routers are configured to work only with the FMS, you are required to forward the stream so you will be able to use it during calibration.
https://docs.limelightvision.io/en/latest/best_practices.html#before-an-event
As a reference

src/main/java/frc/robot/subsystems/vision/Limelight.java
barel13: I think they are using logged subsystem for the limelight because they want to know what is the distance and the angle to the target. Also, it will help them later during the competition after a match to investigate problems

src/main/java/frc/robot/subsystems/vision/Limelight.java
barel13: I think katzuv means that according to the conventions of the team, the name of the function should be `getInstance()`

src/main/java/frc/robot/subsystems/vision/Limelight.java
katzuv: First, use "we" — you are part of the team! 😄
Second, I still don't understand why `LoggedSubsystem` is required. AdvantageKit supports all the logging stuff by itself, it's one of its main features.
I recommend again looking at 6328's code linked above, it should be sufficient.

src/main/java/frc/robot/subsystems/vision/Limelight.java
katzuv: I wanted Emma to answer 🙂.
That is correct though, you can't access the UI of Limelight/PhotonVision when connected to the roboRIO with USB.
Forwarding the UI port through the roboRIO enables us to access the UI regardless of the restrictions Barel mentioned.

I'll also mention that the Limelight docs Barel linked to above say to forward other ports than just `5801`, please do so.

src/main/java/frc/robot/subsystems/vision/Constants.java
```diff
@@ -0,0 +1,9 @@
+package frc.robot.subsystems.vision;
+
+public class Constants {
+    public static final double CAMERA_HEIGHT = 1.1; //[m]
+    public static final double CAMERA_PITCH = -0.611; //[radian]
+    public static final double TARGET_HEIGHT = 0.65; //[m]
+    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0; //[m]
+    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0; //[m]
```
GaiaZano05: Add measurments for Cone x&y.

GaiaZano05 (commented):
Great job!

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -102,16 +103,9 @@ public Optional<Pose2d> estimatePose(Rotation2d robotAngle) {
...
-    /**
-     * updates the limelight log inputs periodically
-     */
-    public void periodic() {
-        updateInputs();
-    }
-
     @Override
     public String getSubsystemName() {
-        return "Limelight";
```
katzuv: Why not call this `Limelight` really? Other vision processing solutions may be used.

src/main/java/frc/robot/subsystems/vision/Constants.java
```diff
@@ -1,8 +1,9 @@
...
 public class Constants {
-    public static final double CAMERA_HEIGHT = 0;
-    public static final double CAMERA_PITCH = 0;
-    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0;
-    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0;
+    public static final double CAMERA_HEIGHT = 1.1; //[m]
+    public static final double CAMERA_PITCH = -0.611; //[radian]
+    public static final double TARGET_HEIGHT = 0.65; //[m]
+    public static final double UPPER_CONE_TARGET11_X_DISTANCE = 0; //[m]
+    public static final double UPPER_CONE_TARGET11_Y_DISTANCE = 0; //[m]
```
katzuv: It's better to represent this information as [`Posed3d`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/geometry/Pose3d.html)s.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,127 @@
...
+    private final IntegerSubscriber tid = table.getIntegerTopic("tid").subscribe(0);
+
+    private Limelight() {
+        super(new LimelightLogInputs());
+        PortForwarder.add(5800, "limelight.local", 5800);
+        PortForwarder.add(5801, "limelight.local", 5801);
+        PortForwarder.add(5802, "limelight.local", 5802);
+        PortForwarder.add(5803, "limelight.local", 5803);
+        PortForwarder.add(5804, "limelight.local", 5804);
+        PortForwarder.add(5805, "limelight.local", 5805);
```
katzuv: There is a lot of repetition in the code block. Find another way to reduce repetition.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,127 @@
...
+        double absouluteAngle = robotAngle.getRadians() + getYaw().getAsDouble();
+        double xTargetDistance = getTargetDistance().getAsDouble() * Math.sin(absouluteAngle);
+        double yTargetDistance = getTargetDistance().getAsDouble() * Math.cos(absouluteAngle);
+        if (hasTargets()) {
+            double xDistance = xTargetDistance - Constants.UPPER_CONE_TARGET11_X_DISTANCE;
+            double yDistance = yTargetDistance - Constants.UPPER_CONE_TARGET11_Y_DISTANCE;
+            Translation2d translation2d = new Translation2d(xDistance, yDistance);
+            return Optional.of(new Pose2d(translation2d, robotAngle));
+        }
+        return Optional.empty();
```
katzuv: It is generally good to reduce indentation (amount of tabs going inside the line) where possible. A popular solution in this case is to invert the condition:
```suggestion
        if (!hasTargets()) {
            return Optional.empty();
        }
        double xDistance = xTargetDistance - Constants.UPPER_CONE_TARGET11_X_DISTANCE;
        double yDistance = yTargetDistance - Constants.UPPER_CONE_TARGET11_Y_DISTANCE;
        Translation2d translation2d = new Translation2d(xDistance, yDistance);
        return Optional.of(new Pose2d(translation2d, robotAngle));
```

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,127 @@
...
+     * @return
+     */
+    public Rotation2d getPitch() {
+        return Rotation2d.fromDegrees(ty.get());
+    }
+
+    /**
+     *returns the distance from the target
+     * @return
+     */
```
katzuv: 1. Sentences start with a capital letter.
2. Fill in the `@return` block tags.

katzuv (changes requested):

katzuv:
I accidentally deleted the `main` branch after merging #6. This deletion caused this PR to be automatically closed, so I'm reopening it. Sorry for the trouble!

src/main/java/frc/robot/subsystems/vision/Constants.java
```diff
@@ -0,0 +1,10 @@
+package frc.robot.subsystems.vision;
+
+import edu.wpi.first.math.geometry.Pose3d;
+import edu.wpi.first.math.geometry.Rotation3d;
+
+public class Constants {
+    public static final double CAMERA_HEIGHT = 1.1; //[m]
+    public static final double CAMERA_PITCH = -0.611; //[radian]
+    public static final Pose3d UPPER_CONE_LEFT_GRID_TARGET = new Pose3d(16.178, 4.99, 1.17, new Rotation3d(0, 0,0)); //[m]
```
katzuv: Note that `Rotation3d` has a [constructor that takes no parameters](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/geometry/Rotation3d.html#%3Cinit%3E()):
> Constructs a Rotation3d with a default angle of 0 degrees.

Also, the unites are incorrect — `m` is a unit of length, but `Pose3d` is not just a single length. However, [IMO](https://dictionary.cambridge.org/dictionary/english/imo) in this case where you use `Pose3d`, you don't have to mention the units because they can be found in the documentation about `Pose3d`.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
+package frc.robot.subsystems.vision;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Pose3d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.net.PortForwarder;
+import edu.wpi.first.networktables.*;
```
katzuv, line 8: It's bad practice to use star `improt`s, because you don't want to `import` _everyhting_ inside `edu.wpi.first.networktables`. If you'll remove this line,
IntelliJ IDEA will show you which `import` statements are missing and offer you to add them.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return INSTANCE;
+    }
+
+    /**
+     * Checks weather the limelight can see any targets
+     * @return weather the limelight detects any targets
+     */
+    public boolean hasTargets() {
+        return tv.get() == 1;
```
katzuv: What if Limelight detects three targets?

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param robotAngle The angle of the robot
+     * @return Optional Pose2d of the robot coordinates
+     */
+    public Optional<Pose2d> estimatePose(Rotation2d robotAngle, Pose3d target) {
+        double absoluteAngle = robotAngle.getRadians() + getYaw().orElse(0);
+        double xTargetDistance = getTargetDistance(target.getZ()).orElse(0) * Math.sin(absoluteAngle);
+        double yTargetDistance = getTargetDistance(target.getZ()).orElse(0) * Math.cos(absoluteAngle);
+        if (!hasTargets()) {
+            return Optional.empty();
+        }
```
katzuv, line 95: This can be moved to the top of the function, there is no reason to calculate `absoluteAngle`, `xTargetDistance` and `yTargetDistance` if they won't be used.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Updates the limelight log inputs
+     */
+    public void updateInputs() {
+        loggerInputs.hasTargets = hasTargets();
+        getYaw().ifPresent((value) -> loggerInputs.yaw = value);
+        loggerInputs.tagId = getTagId();
+        getTargetDistance(117).ifPresent((value) -> loggerInputs.targetDistance = value);
+        getTargetDistance(65).ifPresent((value) -> loggerInputs.targetDistance = value);
```
katzuv: 1. Magic numbers.
2. hese two lines are almost identical. Is there a way to reduce repetition?
Remember, [Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don't_repeat_yourself).

src/main/java/frc/robot/subsystems/vision/LimelightLogInputs.java
```diff
@@ -0,0 +1,37 @@
...
+import org.littletonrobotics.junction.LogTable;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+//@AutoLog
+//public class LimelightLogInputs {
+//    public double targetDistance = 0;
+//    public double yaw = 0;
+//    public boolean hasTargets = false;
+//    public float tagId = 0;
+//}
```
katzuv, line 13: Don't push commented-out code. If you think you'll need it in the future, just commit it and then delete it — Git will save this code in the code history.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+        return Optional.of(new Pose2d(translation2d, robotAngle));
+    }
+
+    @Override
+    public String getSubsystemName() {
+        return "Limelight";
+    }
+
+    /**
+     * Updates the limelight log inputs
```
katzuv:
```suggestion
     * Updates the Limelight log inputs.
```

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+    public Rotation2d getPitch() {
+        return Rotation2d.fromDegrees(ty.get());
+    }
+
+    /**
+     * @return distance from target
+     */
+    public OptionalDouble getTargetDistance(double targetHeight) {
+        double totalPitch = Constants.CAMERA_PITCH + getPitch().getRadians();
+        double error = Constants.CAMERA_HEIGHT-targetHeight;
```
katzuv, line 73: "error" of what? Variable names need to be descriptive and clear, just like you didn't name the previous variable just "total" but "total pitch".

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+        if (hasTargets()) {
+            return OptionalDouble.of(Math.abs(error / Math.tan(totalPitch)));
```
katzuv: Maybe I don't understand something, but shouldn't sin be used instead of tan?

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+    public Optional<Pose2d> estimatePose(Rotation2d robotAngle, Pose3d target) {
+        double absoluteAngle = robotAngle.getRadians() + getYaw().orElse(0);
+        double xTargetDistance = getTargetDistance(target.getZ()).orElse(0) * Math.sin(absoluteAngle);
+        double yTargetDistance = getTargetDistance(target.getZ()).orElse(0) * Math.cos(absoluteAngle);
+        if (!hasTargets()) {
+            return Optional.empty();
+        }
+        double xDistance = xTargetDistance - target.getX();
+        double yDistance = yTargetDistance - target.getY();
+        Translation2d translation2d = new Translation2d(xDistance, yDistance);
```
katzuv: [`Pose3d` has a pretty extensive API](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/geometry/Pose3d.html). Perhaps it contains a method that makes all this work much easier?

katzuv (changes requested):
Getting better! 💯

src/main/java/frc/robot/subsystems/vision/Limelight.java
Emma03L: It was automatically imported by intellij

katzuv: Try deleting this line, when you hover over the lines which need the import (e.g. line 20), IntelliJ IDEA suggests to insert this import.

src/main/java/frc/robot/subsystems/vision/Limelight.java
Emma03L: No its supposed to be tan but the calculation is wrong because I used ty (from the network tables).
The calculation should be the error * tan(camera pitch)

barel13: I think you have a mistake, according to the [documentation](https://docs.limelightvision.io/en/latest/cs_estimating_distance.html#using-a-fixed-angle-camera) of limelight, the code is correct

katzuv: I see, thank you. I imagined the given distance is the hypotenuse of the triangle and not as returned.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.net.PortForwarder;
+import edu.wpi.first.networktables.*;
+import frc.robot.subsystems.LoggedSubsystem;
+
+import java.util.Optional;
+import java.util.OptionalDouble;
+
+public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
+    public static Limelight INSTANCE = null;
```
barel13, line 15:
```suggestion
    private static Limelight INSTANCE = null;
```

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * @return distance from target
+     */
+    public OptionalDouble getTargetDistance(double targetHeight) {
+        if (!hasTargets()) {
+            return OptionalDouble.empty();
+        }
+        double totalPitch = Constants.CAMERA_PITCH + getPitch().getRadians();
+        double error = Constants.CAMERA_HEIGHT-targetHeight;
+        return OptionalDouble.of(Math.abs(error * Math.tan(totalPitch)));
```
barel13, line 74: I know you've changed it back, but just in case, it should be `error / Math.tan(totalPitch)`

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * @return robot yaw
+     */
+    public OptionalDouble getYaw() {
+        if (hasTargets()) {
+            return OptionalDouble.of(ts.get());
+        }
+        return OptionalDouble.empty();
+    }
```
barel13, line 85: It might be more useful to use Rotation2d, so in the future you won't use the wrong units

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Override
+    public String getSubsystemName() {
+        return "Limelight";
+    }
+
+    /**
+     * Updates the limelight log inputs.
+     */
+    public void updateInputs() {
```
barel13, line 113:
```suggestion
    @Override
    public void updateInputs() {
```

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.net.PortForwarder;
+import edu.wpi.first.networktables.*;
+import frc.robot.subsystems.LoggedSubsystem;
+
+import java.util.Optional;
+import java.util.OptionalDouble;
+
+public class Limelight extends LoggedSubsystem<LimelightLogInputs> {
```
barel13, line 14: If I remember correctly, sometimes using networktables, as they add latency, it's more preferable to store the values inside an object, as by the time you pass the check the value of `hasTarget()` can change. (That's why photonvision stores the values in `PhotonPipelineResult`)

src/main/java/frc/robot/subsystems/vision/LimelightLogInputs.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class LimelightLogInputs implements LoggableInputs {
+    public double targetDistance = 0;
+    public double yaw = 0;
+    public boolean hasTargets = false;
+    public float tagId = 0;
+
+    public LimelightLogInputs() {
+    }
```
barel13, line 22: The constructor is unnecessary, Java will automatically generate empty public constructor for you

src/main/java/frc/robot/subsystems/vision/Limelight.java
barel13: (Just for general knowledge, Java doesn't do anything with imported classes that you don't use, it loads them when you first use them, so the * import doesn't have a real affect other than clearance)

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param robotAngle The angle of the robot
+     * @return Optional Pose2d of the robot coordinates
+     */
+    public Optional<Pose2d> estimatePose(Rotation2d robotAngle, Pose3d target) {
+        if (!hasTargets()) {
+            return Optional.empty();
+        }
+        double absoluteAngle = robotAngle.getRadians() + getYaw().orElse(0);
+        double xTargetDistance = getTargetDistance(target.getZ()).orElse(0) * Math.sin(absoluteAngle);
+        double yTargetDistance = getTargetDistance(target.getZ()).orElse(0) * Math.cos(absoluteAngle);
```
barel13, line 98: It's not required, just my personal preference, but if you will use `Rotation2d` for `absoluteAngle` then you can just use the `getCos()/getSin()` functions.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return OptionalDouble.empty();
+    }
+
+    /**
+     * Estimates the position of the robot
+     * @param robotAngle The angle of the robot
+     * @return Optional Pose2d of the robot coordinates
+     */
+    public Optional<Pose2d> estimatePose(Rotation2d robotAngle, Pose3d target) {
```
barel13, line 92: Why not just get the height? as I see, this is the only component you use.

src/main/java/frc/robot/subsystems/vision/Limelight.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Updates the limelight log inputs.
+     */
+    public void updateInputs() {
+        loggerInputs.hasTargets = hasTargets();
+        getYaw().ifPresent((value) -> loggerInputs.yaw = value);
+        loggerInputs.tagId = getTagId();
+        getTargetDistance(Constants.UPPER_CONE_TARGET_TAPE_HEIGHT).ifPresent((value) -> loggerInputs.targetDistance = value);
+        getTargetDistance(Constants.LOWER_CONE_TARGET_TAPE_HEIGHT).ifPresent((value) -> loggerInputs.targetDistance = value);
```
barel13, line 118: You should know which target you are tracking now. Using this code, if you see a target, the lower cone target value will always override the value of the upper cone

src/main/java/frc/robot/subsystems/vision/LimelightLogInputs.java
```diff
@@ -0,0 +1,37 @@
...
+//    public double yaw = 0;
+//    public boolean hasTargets = false;
+//    public float tagId = 0;
+//}
+
+public class LimelightLogInputs implements LoggableInputs {
+    public double targetDistance = 0;
+    public double yaw = 0;
+    public boolean hasTargets = false;
+    public float tagId = 0;
```
barel13, line 19: float?

barel13 (changes requested):

src/main/java/frc/robot/subsystems/vision/Limelight.java
katzuv: Interesting, I did not know that.
It's still bad practice to use imports of this kind, they are less clear.

src/main/java/frc/robot/subsystems/vision/Limelight.java
vichik123: I know that there is a new `AutoLog` annotation that 6328 released. We haven't had the opportunity to change it yet.

vichik123 (approved):
Looks pretty good!

katzuv:
It's a very quick change, just delete `toLog` and `fromLog`, add the annotation, and you're done.

PR #7, opened 2023-01-19
Create prototypeArm subsystem

Emma03L:
Use 2 mini CIM motors and 2 MAG encoders to control the position of the 2 jointed arm.


src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
```diff
@@ -0,0 +1,20 @@
...
+    public SetElbowAngle(PrototypeArm prototypeArm, double angle) {
+        this.prototypeArm = prototypeArm;
+        this.angle = angle;
+        addRequirements(prototypeArm);
+    }
+
+    @Override
+    public void execute() {
+        prototypeArm.setElbowJointPosition(angle);
+    }
```
GaiaZano05, line 19: Where is the end?

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,26 @@
...
+        this.prototypeArm = prototypeArm;
+        this.position = position;
+        addRequirements();
+    }
+
+    @Override
+    public void execute() {
+        prototypeArm.setPosition(position);
+    }
+}
```
GaiaZano05, line 27: Where is the end?

src/main/java/frc/robot/subsystems/arm/commands/SetShoulderAngle.java
```diff
@@ -0,0 +1,20 @@
...
+        this.prototypeArm = prototypeArm;
+        this.angle = angle;
+        addRequirements(prototypeArm);
+    }
+
+    @Override
+    public void execute() {
+        prototypeArm.setShoulderJointPosition(angle);
+    }
+}
```
GaiaZano05, line 26: Where is the end?

src/main/java/frc/robot/Ports.java
```diff
@@ -2,4 +2,11 @@
 
 public final class Ports {
 
+    public static class prototypeArmPorts{
+        public static final int SHOULDER_MOTOR = 0;
+        public static final int ELBOW_MOTOR = 0;
+        public static final int SHOULDER_ENCODER = 0;
+        public static final int ELBOW_ENCODER = 0;
+    }
```
katzuv: This needs to be in the subsystem-specific ports file, not here (same as with the vision constants).

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final boolean clockWise = true;
+
+    public static final double kP = 0.0;
+    public static final double kI = 0.0;
+    public static final double kD = 0.0;
+
+    public static final double ENCODER_RATIO = 1/25;
+    public static final double TICKS_PER_RADIAN = (4096/ENCODER_RATIO)/ (Math.PI*2);
+
+    public static final double SHOULDER_GEARING = (66/16)*25; // Arbitrary units
```
katzuv: Wikipedia says the term is ["dimensionless"](https://en.wikipedia.org/wiki/Dimensionless_physical_constant),

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final double SHOULDER_MOMENT_OF_INERTIA = 0.0212405825*0.9; // [kg*m^2]
+    public static final double SHOULDER_CM_RADIUS = 0.15869; // [m]
+    public static final int SHOULDER_NUMBER_OF_MOTORS = 1; // Arbitrary units
+    public static final double SHOULDER_ARM_LENGTH = 0.385; //[m]
+
+    public static final double ELBOW_GEARING = (66/16)*25; // Arbitrary units
+    public static final double ELBOW_MASS =0.2859; // [kg]
+    public static final double ELBOW_LENGTH = 0.385; // [m]
+    public static final double ELBOW_MOMENT_OF_INERTIA = 0.0212405825*0.9; // [kg*m^2]
+    public static final double ELBOW_CM_RADIUS = 0.15869; // [m]
```
katzuv, line 32: 1. We usually don't include units in names.
2. The constant name contains "cm", but the comment says it's in meters.

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
+package frc.robot.subsystems.arm;
+
+import edu.wpi.first.math.geometry.Translation2d;
+
+/**
+ * This class contains the kinematics for the arm.
+ * All the calculations are carefully documented in the following page:
```
katzuv:
```suggestion
 * All the calculations are documented on the following page:
```

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
...
+ * https://www.societyofrobots.com/robot_arm_tutorial.shtml#inverse_kinematics
+ * There is a table of contents containing the kinematics section, and the inverse kinematics section.
+ */
+public class ArmKinematics {
+    private final double l1;
+    private final double l2;
```
katzuv: Don't shorten — use the full form of the word, especially when `l` looks very similar to `1`.

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Inverse kinematics for the arm.
+     *
+     * @param x the x coordinate of the end effector. [m]
+     * @param y the y coordinate of the end effector. [m]
+     * @return the angles of the shoulder and elbow joints. ([rad, rad])
+     */
+    public InverseKinematicsSolution inverseKinematics(double x, double y) {
```
katzuv, line 38: It's better to stay consistent — if `forwardKinematics()` returns a `Translation2d`, it's good that this function will receive a parameter of type `Translation2d`.

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
...
+        double theta = Math.acos((y * (l1 + l2 * c2) - x * l2 * s2) / (x * x + y * y));
+        double elbowAngle = theta - shoulderAngle + Math.PI / 2;
+
+        return new InverseKinematicsSolution(shoulderAngle, elbowAngle);
+    }
+
+    /**
+     * A class representing the solution of the inverse kinematics.
+     */
+    public static class InverseKinematicsSolution {
```
katzuv, line 59: This class can be ~private~ protected.

src/main/java/frc/robot/utils/motors/FalconConstants.java
```diff
@@ -0,0 +1,11 @@
+package frc.robot.utils.motors;
+
+public class FalconConstants {
+    public static final double NOMINAL_VOLTAGE = 12.0; // [V]
```
katzuv, line 4: This constant is relevant to stuff other than Falcon 500 motors, so it shouldn't be in this class.

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
...
+        double y = l1 * Math.sin(shoulderAngle) + l2 * Math.cos(theta);
+        return new Translation2d(x, y);
+    }
+
+    /**
+     * Inverse kinematics for the arm.
+     *
+     * @param x the x coordinate of the end effector. [m]
+     * @param y the y coordinate of the end effector. [m]
+     * @return the angles of the shoulder and elbow joints. ([rad, rad])
```
katzuv: I suggest using [`Rotation2d`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/geometry/Rotation2d.html) here. two advantages:
- No need to use radians. Degrees are easier for end-users to work with — it's easier to say "change the second arm's angle to 60°" instead of "change the second arm's angle to 1.05rad".
- You will be able to simply do `shoulderAngle.getSin()` instead of `Math.sin(shoulderAngle)`.

src/main/java/frc/robot/Ports.java
```diff
@@ -2,4 +2,11 @@
 
 public final class Ports {
 
+    public static class prototypeArmPorts{
+        public static final int SHOULDER_MOTOR = 22;
+        public static final int ELBOW_MOTOR = 21;
+        public static final int SHOULDER_ENCODER = 22;
+        public static final int ELBOW_ENCODER = 21;
+    }
```
katzuv, line 10: Why is this in the global ports file?

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,42 @@
+package frc.robot.subsystems.arm;
+
+public class ArmConstants { // TODO:
```
katzuv: Empty TODO.

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,42 @@
...
+
+    // Arm constants
+    public static final double CONFIG_VOLT_COMP = 10; //[voltage]
```
katzuv, line 6:
```suggestion
    public static final double CONFIG_VOLT_COMP = 10; //[V]
```
Also:
1. The word "config" is redundant, it's obvious this constant is used for configuration.
2. Don't shorten words.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+
+import com.ctre.phoenix.motorcontrol.*;
+import com.ctre.phoenix.motorcontrol.can.TalonSRX;
+import com.ctre.phoenix.sensors.CANCoder;
+import edu.wpi.first.math.geometry.Translation2d;
+import frc.robot.Ports;
+import frc.robot.subsystems.LoggedSubsystem;
+import frc.robot.utils.units.UnitModel;
+
+public class PrototypeArm extends LoggedSubsystem<PrototypeArmLogInputs> {
```
katzuv, line 15: Why do you need `LoggedSubsystem`?

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+        timer.start();
+
+        shoulderMotor.configVoltageCompSaturation(ArmConstants.CONFIG_VOLT_COMP);
+        shoulderMotor.enableVoltageCompensation(ArmConstants.ENABLE_VOLT_COMPANSATION);
+        shoulderMotor.setNeutralMode(NeutralMode.Brake);
+        shoulderMotor.setInverted(ArmConstants.clockWise);
+        shoulderMotor.config_kP(0, ArmConstants.shoulderP);
+        shoulderMotor.config_kI(0, ArmConstants.shoulderI);
+        shoulderMotor.config_kD(0, ArmConstants.shoulderD);
+//        shoulderMotor.configSelectedFeedbackSensor(TalonSRXFeedbackDevice.Analog, 0, 20);
```
katzuv: Don't leave commented-out code in the repository. It is now safely saved in Git. You can delete it now, and if you'll need it in the future, just come back to the relevant commit and get the code back.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+import edu.wpi.first.math.geometry.Translation2d;
+import frc.robot.Ports;
+import frc.robot.subsystems.LoggedSubsystem;
+import frc.robot.utils.units.UnitModel;
+
+public class PrototypeArm extends LoggedSubsystem<PrototypeArmLogInputs> {
+    private final ArmKinematics kinematics = new ArmKinematics(ArmConstants.SHOULDER_ARM_LENGTH, ArmConstants.ELBOW_ARM_LENGTH);
+    private final ArmSystemModel systemModel = new ArmSystemModel(ArmConstants.ARM_CONSTANTS);
+
+    public final edu.wpi.first.wpilibj.Timer timer = new edu.wpi.first.wpilibj.Timer();
```
katzuv: Fix the imports so that line will look line:
```suggestion
    public final Timer timer = new Timer();
```

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+        shoulderMotor.set(ControlMode.MotionMagic, unitModel.toTicks(Math.toRadians(angle)), DemandType.ArbitraryFeedForward, shoulderFeedforward);
+    }
+
+    public double getElbowJointPosition() {
+        return Math.toDegrees(unitModel.toUnits(elbowMotor.getSelectedSensorPosition()));
+    }
+
+    public void setElbowJointPosition(double angle) {
+        elbowMotor.set(ControlMode.MotionMagic, unitModel.toTicks(Math.toRadians(angle)), DemandType.ArbitraryFeedForward, elbowFeedForward);
+    }
```
katzuv: I'd replace "position" with "angle", it's clearer.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+        return shoulderMotor.getSelectedSensorVelocity();
+    }
+
+    public double getElbowMotorVelocity(){
+        return elbowMotor.getSelectedSensorVelocity();
+    }
+
+    public void setPosition(Translation2d armLocation){
+        setShoulderJointPosition(kinematics.inverseKinematics(armLocation.getX(), armLocation.getY()).shoulderAngle);
+        setElbowJointPosition(kinematics.inverseKinematics(armLocation.getX(), armLocation.getY()).elbowAngle);
```
katzuv, line 121: Remember, Don't Repeat Yourself:
```suggestion
        var angles = kinematics.inverseKinematics(armLocation.getX(), armLocation.getY());
        setShoulderJointPosition(angles.shoulderAngle);
        setElbowJointPosition(angles.elbowAngle);

File: src/main/java/frc/robot/subsystems/arm/PrototypeArm.java [hunk 21]
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    public double deadBend(double value){
+        if (Math.abs(value)>0.05) return value;
+        return 0;
+    }
```
katzuv: The CTRE API supports deadband out of the box. Read [here](https://docs.ctre-phoenix.com/en/stable/ch13_MC.html#neutral-deadband) for more information.
**Note**: the documentation says the default headband is 4%, which may be enough.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public String getSubsystemName() {
+        return "PrototypeArm";
+    }
+
+    public void periodic(){
+        timer.reset();
+        timer.start();
```
katzuv, line 131: Do you really want to reset the timer periodically?

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public final TalonSRX shoulderMotor = new TalonSRX(Ports.prototypeArmPorts.SHOULDER_MOTOR);
+    public final TalonSRX elbowMotor = new TalonSRX(Ports.prototypeArmPorts.ELBOW_MOTOR);
+    public final UnitModel unitModel = new UnitModel(ArmConstants.TICKS_PER_RADIAN);
+
+    private double prevShoulderVelocity;
+    private double shoulderFeedforward;
+    private double prevElbowVelocity;
+    private double elbowFeedForward;
+    private double Time2;
```
katzuv: Variable names are written in `lowerCamelCase`.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+        timer.reset();
+        timer.start();
+        double shoulderVelocity = getShoulderMotorVelocity();
+        double elbowVelocity = getElbowMotorVelocity();
+        double Time = timer.get();
+        double shoulderAcceleration = (shoulderVelocity - prevShoulderVelocity)/ Math.abs(Time-Time2);
+        double elbowAcceleration = (elbowVelocity - prevElbowVelocity)/Math.abs(Time-Time2);
+        prevShoulderVelocity = shoulderVelocity;
+        prevElbowVelocity = elbowVelocity;
+        Time2 = timer.get();
```
katzuv: You don't need the timer object. You can use [`Timer.getFPGATimestamp()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj/Timer.html#getFPGATimestamp()) directly;
```suggestion
        double Time = Timer.getFPGATimestamp();
        double shoulderAcceleration = (shoulderVelocity - prevShoulderVelocity)/ Math.abs(Time-Time2);
        double elbowAcceleration = (elbowVelocity - prevElbowVelocity)/Math.abs(Time-Time2);
        prevShoulderVelocity = shoulderVelocity;
        prevElbowVelocity = elbowVelocity;
        Time2 = Timer.getFPGATimestamp();
```
_However_...
You probably don't want to set both `Time` and `Time2` here — the time which passes between executing lines 120 and 125 is very, very short. IIUC, you should only update the "previous timestamp" with the "current timestamp", then update "current timestamp" with `Timer.getFPGATimestamp()`.
Last thing: "time" and "time 2" are not very good names — it's unclear which of them is earlier.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+
+        ArmSystemModel.ArmFeedForward tempFeedforward = systemModel.calculateFeedForward(Math.toDegrees(unitModel.toUnits(getShoulderJointPosition())), Math.toDegrees(unitModel.toUnits(getElbowJointPosition())), getShoulderMotorVelocity(), getElbowMotorVelocity(), shoulderAcceleration, elbowAcceleration);
+        shoulderFeedforward = tempFeedforward.shoulderFeedForward;
+        elbowFeedForward = tempFeedforward.elbowFeedForward;
```
katzuv: It's probably better to declare these variable at the class level, instead of creating them again each loop.

src/main/java/frc/robot/subsystems/arm/SystemConstants.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public JointConstants getElbowJointConstants() {
+        return elbowJointConstants;
+    }
+
+    public static class JointConstants {
+        public double mass;
+        public double length;
+        public double momentOfInertia;
+        public double cmRadius;
```
katzuv, line 30: The radius of what? Also, it's better to write "radius cm" instead of the other way around.

src/main/java/frc/robot/subsystems/arm/SystemConstants.java
```diff
@@ -0,0 +1,43 @@
...
+    private final JointConstants shoulderJointConstants;
+    private final JointConstants elbowJointConstants;
+
+    public SystemConstants(JointConstants shoulderJointConstants, JointConstants elbowJointConstants) {
+        this.shoulderJointConstants = shoulderJointConstants;
+        this.elbowJointConstants = elbowJointConstants;
+    }
+
+    public SystemConstants(double shoulderMass, double shoulderLength, double shoulderMomentOfInertia, double shoulderCmRadius, double shoulderGearing, int shoulderNumMotors,
+                           double elbowMass, double elbowLength, double elbowMomentOfInertia, double elbowCmRadius, double elbowGearing, int elbowNumMotors) {
```
katzuv, line 13: `elbowRadiusCm` is better than `elbowCmRadius`.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public SetArmsPosition(PrototypeArm prototypeArm, Translation2d position) {
+        this.prototypeArm = prototypeArm;
+        this.position = position;
+        addRequirements();
+    }
+
+    @Override
+    public void execute() {
+        prototypeArm.setPosition(position);
```
katzuv, line 19: It's unnecessary to set the setpoint in every loop. You can set the setpoint once at `initialize()` and the Talon will "remember" it until you set it to something else.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
katzuv: IIUC you _don't_ want to set the powers to zero, because then the motor won't actively hold the subsystem in place.
On the other end, `isFinished()` does seem like it's missing here.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,32 @@
...
+import frc.robot.subsystems.arm.ArmConstants;
+import frc.robot.subsystems.arm.ArmKinematics;
+import frc.robot.subsystems.arm.ArmSystemModel;
+import frc.robot.subsystems.arm.PrototypeArm;
+
+public class SetArmsPosition extends CommandBase {
+    private final PrototypeArm prototypeArm;
+    private final Translation2d position;
+
+    public SetArmsPosition(PrototypeArm prototypeArm, Translation2d position) {
```
katzuv: Is it necessary to have seperate command for each part of the arm?

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+
+    public double getShoulderMotorPower() {
+        return shoulderMotor.getMotorOutputPercent();
+    }
+
+    public double getElbowMotorPower() {
+        return elbowMotor.getMotorOutputPercent();
+    }
```
katzuv, line 109: What is a use case for these functions?

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
katzuv: Note that after testing is done, all of the setters to specific joints should be private. End users should control the whole arm, not the specific joints it consists of.

src/main/java/frc/robot/utils/motors/FalconConstants.java
```diff
@@ -0,0 +1,11 @@
...
+    public static final double STALL_CURRENT = 783; // [A]
+    public static final double STALL_TORQUE = 4.69; // [N*m]
+    public static final double FREE_SPEED = (6380 / 60.0) * (2 * Math.PI); // [rad/s]
+    public static final double Kt = STALL_TORQUE / STALL_CURRENT; // [N*m/A]
+    public static final double Kv = FREE_SPEED / NOMINAL_VOLTAGE; // [rad/s/V]
+    public static final double R = NOMINAL_VOLTAGE / STALL_CURRENT; // [Ohm]
```
katzuv, line 10: Don't forget that on the real robot, these gains should be derived using SysId.

katzuv, line 10: Don't forget that on the real robot, these gains should be derived using SysId.

src/main/java/frc/robot/utils/motors/FalconConstants.java
```diff
@@ -0,0 +1,11 @@
+package frc.robot.utils.motors;
+
+public class FalconConstants {
```
katzuv, line 3: 1. Isn't a MiniCIM used in the prototype?
2. You can simply use [`DCMotor.getMiniCIM()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/system/plant/DCMotor.html#getMiniCIM(int)) for all of these data.

katzuv (changes requested):
Great job! Hopefully, this prototype will become a full-scale arm in the near future 🤩

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final boolean ENABLE_VOLT_COMPANSATION = true;
```
barel13: I'm sorry but
```suggestion
    public static final boolean ENABLE_VOLT_COMPENSATION = true;
```

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final boolean clockWise = true;
```
barel13, line 8: I don't know if the new convention is to set the inversion in the constant class, but last year we put it in `Ports`. (@GaiaZano05)

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
barel13: But she uses it in other classes.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+
+public class PrototypeArm extends LoggedSubsystem<PrototypeArmLogInputs> {
+    private final ArmKinematics kinematics = new ArmKinematics(ArmConstants.SHOULDER_ARM_LENGTH, ArmConstants.ELBOW_ARM_LENGTH);
+    private final ArmSystemModel systemModel = new ArmSystemModel(ArmConstants.ARM_CONSTANTS);
+
+    public final edu.wpi.first.wpilibj.Timer timer = new edu.wpi.first.wpilibj.Timer();
+
+    public final TalonSRX shoulderMotor = new TalonSRX(Ports.prototypeArmPorts.SHOULDER_MOTOR);
+    public final TalonSRX elbowMotor = new TalonSRX(Ports.prototypeArmPorts.ELBOW_MOTOR);
+    public final UnitModel unitModel = new UnitModel(ArmConstants.TICKS_PER_RADIAN);
```
barel13:
```suggestion
    private final edu.wpi.first.wpilibj.Timer timer = new edu.wpi.first.wpilibj.Timer();

    private final TalonSRX shoulderMotor = new TalonSRX(Ports.prototypeArmPorts.SHOULDER_MOTOR);
    private final TalonSRX elbowMotor = new TalonSRX(Ports.prototypeArmPorts.ELBOW_MOTOR);
    private final UnitModel unitModel = new UnitModel(ArmConstants.TICKS_PER_RADIAN);
```

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private double prevShoulderVelocity;
+    private double shoulderFeedforward;
+    private double prevElbowVelocity;
+    private double elbowFeedForward;
+    private double Time2;
+
+    public PrototypeArm() {
```
barel13: Please make the constructor private and add `getInstance()` function

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
```diff
@@ -0,0 +1,145 @@
...
+        return elbowMotor.getMotorOutputPercent();
+    }
+
+    public double getShoulderMotorVelocity(){
+        return shoulderMotor.getSelectedSensorVelocity();
+    }
+
+    public double getElbowMotorVelocity(){
+        return elbowMotor.getSelectedSensorVelocity();
+    }
```
barel13: Convert the values using the unitModel

src/main/java/frc/robot/subsystems/arm/PrototypeArmLogInputs.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public void toLog(LogTable table) {
+        table.put("elbowAngle", elbowAngle);
+        table.put("shoulderAngle", shoulderAngle);
+        table.put("elbowMotorPower", elbowMotorPower);
+        table.put("shoulderMotorPower", shoulderMotorPower);
+        table.put("shoulderP", shoulderP);
+        table.put("shoulderI", shoulderI);
+        table.put("elbowD", elbowD);
```
barel13:
```suggestion
        table.put("shoulderD", shoulderD);
```

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,32 @@
...
+        this.prototypeArm = prototypeArm;
+        this.position = position;
+        addRequirements();
```
barel13, line 14: 👀

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
barel13, line 19: I saw that you only use degrees in your code, so why not make ticks_per_radian ticks_per_degree?

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
barel13, line 15: I see that both parts of arm have the same kind of computation and functions, maybe consider creating a class for each part of the system? (Like in the swerve)

barel13 (changes requested):

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
katzuv: Where? These are all the usages I've found, and they are all in `ArmKinematics.java`:

image

barel13: Here:
image

katzuv: Fair enough, edited my reply.

barel13: It's still should be public, it's being accessed from another class (even another file)

katzuv: Protected classes can be accessed by classes inside the same package. Outside `ArmKinematics.java` it's used only in `PrototypeArm.java`, and both of them are in the `arm` package.

barel13: So package private should be used for this, or just public in my opinion, as there shouldn't be a restriction about the inheritance of this class

katzuv: How do you make a class package-private?
I think the best solution is to write no modifier at all. That would allow the class to be accessed inside the package only and not be subclassed ([source](https://stackoverflow.com/a/215505/10015776)).
I think this is the best solution because `InverseKinematicsSolution` shouldn't be accessed from outside the `arm` package, it's part of the control logic of the arm that shouldn't be relevant to end-users.

barel13: (Package private is writing without a modifier)

katzuv: 👍🏻 That's what I thought.

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
Emma03L: Eitan is the one who named all of these constants,
I'm pretty sure the cm stands for center of mass but I agree it's a bit confusing

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
Emma03L: I think it's mainly for testing, Gaya asked me to add them

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
Emma03L: It's used in the commands to set the motor power to zero when the command is finished.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
Emma03L: I'm pretty sure that if you set the motors to break the arm won't fall

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
Emma03L: I've always put it in the constants

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
Emma03L: They're used in the logger

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
katzuv: > Eitan is the one who named all of these constants

There is no need to blame anyone, it's not very constructive. It's not like I'm mad or something, just found something that could be better.
Anyway, fixing the issue will be good, unless you disagree with the comment.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
katzuv: It's possible that's the case in the prototype because the arms are so light, but I don't imagine this will be the case in the real arm.
Anyway, I'd not set the power to zero, it doesn't "cost" you anything (and the opposite does "cost" you an additional line of code).

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
katzuv: IMO creating a function, that is used only once, and for this purpose, is a bit overkill.
Writing `loggerInputs.elbowMotorPower = elbowMotor.getMotorOutputPercent()` in the `updateInputs` LGTM.

src/main/java/frc/robot/utils/motors/FalconConstants.java
GaiaZano05: Move it to  shared Constants file.

src/main/java/frc/robot/Ports.java
GaiaZano05: We decided to use the same global port file.

src/main/java/frc/robot/Ports.java
```diff
@@ -2,4 +2,11 @@
 
 public final class Ports {
 
+    public static class prototypeArmPorts{
+        public static final int SHOULDER_MOTOR = 22;
+        public static final int ELBOW_MOTOR = 21;
+        public static final int SHOULDER_ENCODER = 22;
+        public static final int ELBOW_ENCODER = 21;
```
barel13, line 9: The encoders are connected through the controller, so you don't need these constants

src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
barel13: Also, there should be isFinished method

barel13 (changes requested):

src/main/java/frc/robot/utils/motors/FalconConstants.java
katzuv: It can be deleted because it's not going to be used, and if you do need it, just use the `DCMotor` class (look it up in the API reference).

src/main/java/frc/robot/Ports.java
katzuv: Before kickoff we had the opposite decision. What is the benefit of having a global ports file? Also, are configurations stored in it too?
The big disadvantage is that many people working on the same file leads to bad merge conflicts.
And @barel13, please don't resolve conversations you didn't open, especially ones you don't participate in…

GaiaZano05: I thought we were tailking only about constants file. I think it's better to leave the ports file as is, beacause it's not as messy as constants file and it's more convinient to have all the ports at the same place.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
GaiaZano05: Add what requirments?

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
GaiaZano05: Yeah I prefer it being in Constants.

src/main/java/frc/robot/Ports.java
katzuv: The high risk for merge conflicts is still relevant, even if the conflicts will be less hard to deal with.

katzuv (changes requested):

katzuv:
Because this is the prototype branch I won't be reviewing this code, all my comments are in #12. If there is specific stuff you want me to look at in this PR, tell me.

src/main/java/frc/robot/subsystems/arm/PrototypeArm.java
Emma03L: But you need the timer object to start it

katzuv: I'm not sure I understood what you meant, but if I did, notice that I used `Timer.getFPGATimestamp()` with a capital T. This is because `getFPGATimestamp()` is a static method, so you don't need to instantiate a `Timer` object to use it.

PR #11, opened 2023-01-22
Feature/grippe

AlonSchwierz:


src/main/java/frc/robot/subsystems/gripper/Commands/toggleSolenoid.java
```diff
@@ -0,0 +1,18 @@
+package frc.robot.subsystems.gripper.Commands;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.subsystems.gripper.Gripper;
+
+public class toggleSolenoid extends CommandBase {
+    private Gripper gripper;
```
barel13: Please initialize the gripper in the constructor and `addRequirements`

src/main/java/frc/robot/subsystems/gripper/Commands/toggleSolenoid.java
```diff
@@ -0,0 +1,18 @@
+package frc.robot.subsystems.gripper.Commands;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.subsystems.gripper.Gripper;
+
+public class toggleSolenoid extends CommandBase {
```
barel13: Should be instant command

src/main/java/frc/robot/subsystems/gripper/Gripper.java
```diff
@@ -0,0 +1,73 @@
...
+import edu.wpi.first.networktables.NetworkTable;
+import edu.wpi.first.networktables.NetworkTableInstance;
+import edu.wpi.first.wpilibj.PneumaticsModuleType;
+import edu.wpi.first.wpilibj.Solenoid;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Ports;
+import frc.robot.subsystems.LoggedSubsystem;
+
+public class Gripper extends LoggedSubsystem<GripperLoggedInputs> {
+    private GripperLoggedInputs gripperLoggedInputs;
```
barel13: There is no need for this, you have thus from the super class

src/main/java/frc/robot/subsystems/gripper/Gripper.java
```diff
@@ -0,0 +1,73 @@
...
+    private final Solenoid leftSolenoid = new Solenoid(PneumaticsModuleType.CTREPCM, Ports.Gripper.LEFT_SOLENOID);
+    private final Solenoid rightSolenoid = new Solenoid(PneumaticsModuleType.CTREPCM, Ports.Gripper.RIGHT_SOLENOID);
+
+    private final NetworkTable table = NetworkTableInstance.getDefault().getTable("gripper");
```
barel13: Well... you don't use this class

barel13 (changes requested):

src/main/java/frc/robot/Ports.java
```diff
@@ -1,5 +1,8 @@
 package frc.robot;
 
 public final class Ports {
-
+public class Gripper{
+    public static final int LEFT_SOLENOID = 0;
+    public static final int RIGHT_SOLENOID = 0;
+}
```
katzuv, line 7: These should be in a ports file in the gripper package.

src/main/java/frc/robot/subsystems/gripper/Gripper.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * return the current state of the solenoids
+     */
+    public void getRightSolenoid() {
+        rightSolenoid.get();
+    }
+
+    public void getLeftSolenoid() {
+        leftSolenoid.get();
+    }
```
katzuv, line 61: What is a use case for these methods?

src/main/java/frc/robot/subsystems/gripper/Gripper.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * changes the current state to the other state
+     */
+
+    public void toggleSolenoid() {
+        leftSolenoid.toggle();
+        rightSolenoid.toggle();
+
+    }
```
katzuv: Implementation details should stay hidden. That means that end-users don't need to know that the gripper is using solenoids.
So, these methods should be called something similar to "open"/"close" and "toggle".

src/main/java/frc/robot/subsystems/gripper/Commands/toggleSolenoid.java
katzuv: The whole class can be replace with an inline command: `new InstantCommand(gripper::toggleSolenoid)`.

katzuv (changes requested):
Nice! A few notes:
- "grippe" in the branch name should be fixed to "gripper". Do this through GitHub.
- The PR's title should be human-readable, not the same as the branch name.
- The `Gripper` class lacks a `getInstance()` method.
- I'd add a "close" method as well to the `Gripper` class.

katzuv:
Also, note that the assignees' list contains the people who _work_ on the PR -- this is you (@AlonSchwierz). If you want a review from someone, just add them to the reviewers' list 😊.

src/main/java/frc/robot/Ports.java
GaiaZano05: We decided to have a shared Port file and a seperate Constants file.

src/main/java/frc/robot/subsystems/gripper/Commands/toggleSolenoid.java
AlonSchwierz: done.

src/main/java/frc/robot/subsystems/gripper/Commands/toggleSolenoid.java
AlonSchwierz: done.

src/main/java/frc/robot/subsystems/gripper/Gripper.java
AlonSchwierz: done.

src/main/java/frc/robot/subsystems/gripper/Gripper.java
AlonSchwierz: OK.

src/main/java/frc/robot/subsystems/gripper/Gripper.java
AlonSchwierz: Well, the only case I can think of is when the arm engages in order to pick up cones, if we want to do that we should know whether the gripper is closed or open.

src/main/java/frc/robot/subsystems/gripper/Gripper.java
AlonSchwierz: done

src/main/java/frc/robot/subsystems/gripper/Gripper.java
vichik123: You get a parameter `GripperLogInputs` in the constructor, and pass it to the super constructor. However, the logger inputs should be instantiated right in the constructor, without being passed as a parameter.

src/main/java/frc/robot/subsystems/gripper/GripperLoggedInputs.java
```diff
@@ -0,0 +1,28 @@
+package frc.robot.subsystems.gripper;
+
+import org.littletonrobotics.junction.LogTable;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+public class GripperLoggedInputs implements LoggableInputs {
+    boolean leftSolenoid;
+    boolean rightSolenoid;
```
vichik123, line 8: Maybe you should give more descriptive names for these fields, like - `leftSolenoidExtended`.

vichik123 (changes requested):

PR #12, opened 2023-01-23
Add non prototype arm branch

Emma03L:
Use 4 falcon motors, 2 for each joint, and 2 talonSRX mag encoders.


src/main/java/frc/robot/Ports.java
```diff
@@ -2,4 +2,13 @@
...
 public final class Ports {
 
+    public static class ArmPorts {
+        public static final int MAIN_SHOULDER_MOTOR = 0;
+        public static final int AUX_SHOULDER_MOTOR = 0;
+        public static final int MAIN_ELBOW_MOTOR = 0;
+        public static final int AUX_ELBOW_MOTOR = 0;
+        public static final int SHOULDER_ENCODER = 0;
+        public static final int ELBOW_ENCODER = 0;
+    }
```
katzuv: These should be in the `arm` package.

src/main/java/frc/robot/Ports.java
```diff
@@ -2,4 +2,13 @@
 
 public final class Ports {
 
+    public static class ArmPorts {
+        public static final int MAIN_SHOULDER_MOTOR = 0;
+        public static final int AUX_SHOULDER_MOTOR = 0;
+        public static final int MAIN_ELBOW_MOTOR = 0;
+        public static final int AUX_ELBOW_MOTOR = 0;
+        public static final int SHOULDER_ENCODER = 0;
+        public static final int ELBOW_ENCODER = 0;
```
katzuv: I think it's better to write "shoulder main motor", then "should aux motor", etc. That's because the motors are a part of the shoulder/elbow, not the other way around.
But anyway, IMO "elbow/shoulder" is not very clear, and "top/bottom" is clearer.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+        mainElbowMotor.setNeutralMode(NeutralMode.Brake);
+        mainElbowMotor.setInverted(ArmConstants.clockWise);
+        mainElbowMotor.config_kP(0, ArmConstants.elbowP);
+        mainElbowMotor.config_kI(0, ArmConstants.elbowI);
+        mainElbowMotor.config_kD(0, ArmConstants.elbowD);
+        mainElbowMotor.configNeutralDeadband(0.05);
+        auxElbowMotor.follow(mainElbowMotor);
+        auxElbowMotor.configVoltageCompSaturation(ArmConstants.VOLT_COMP_Saturation);
+        auxElbowMotor.enableVoltageCompensation(ArmConstants.ENABLE_VOLT_COMPENSATION);
+        auxElbowMotor.setNeutralMode(NeutralMode.Brake);
```
katzuv: Use the `TalonSRXConfiguration` class and the `TalonSRX.configAllSettings()` method. This will cut your configuration code by half because you won't need to call the configuration method for each configuration "detail".

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+        mainElbowMotor.set(TalonFXControlMode.Velocity, velocity);
+    }
+
+    public void setShoulderJointAngle(double angle) {
+        mainShoulderMotor.set(TalonFXControlMode.MotionMagic, unitModel.toTicks(Math.toRadians(angle)), DemandType.ArbitraryFeedForward, shoulderFeedforward);
+    }
+
+    public void setElbowJointAngle(double angle) {
+        mainElbowMotor.set(TalonFXControlMode.MotionMagic, unitModel.toTicks(Math.toRadians(angle)), DemandType.ArbitraryFeedForward, elbowFeedForward);
+    }
```
katzuv: Will some of these functions be deleted in the future? Looks too much to have three different control methods.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public void setAngle(Translation2d armLocation) {
```
katzuv: This name is configuring, it implies this method receives an angle and not a location.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+        setElbowJointAngle(angles.elbowAngle);
+    }
+
+    public double getShoulderJointAngle() {
+        return Math.toDegrees(unitModel.toUnits(shoulderEncoder.getSelectedSensorPosition()));
+    }
+
+    public double getElbowJointAngle() {
+        return Math.toDegrees(unitModel.toUnits(elbowEncoder.getSelectedSensorPosition()));
+    }
```
katzuv: Perhaps these methods should return a `Rotation2d` (not sure what I prefer).

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+        auxShoulderMotor.setNeutralMode(NeutralMode.Brake);
+
+        mainElbowMotor.configVoltageCompSaturation(ArmConstants.VOLT_COMP_Saturation);
+        mainElbowMotor.enableVoltageCompensation(ArmConstants.ENABLE_VOLT_COMPENSATION);
+        mainElbowMotor.setNeutralMode(NeutralMode.Brake);
+        mainElbowMotor.setInverted(ArmConstants.clockWise);
+        mainElbowMotor.config_kP(0, ArmConstants.elbowP);
+        mainElbowMotor.config_kI(0, ArmConstants.elbowI);
+        mainElbowMotor.config_kD(0, ArmConstants.elbowD);
+        mainElbowMotor.configNeutralDeadband(0.05);
```
katzuv: If all other configuration values are stored as constants, this one should too.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+        shoulderVelocity = getShoulderMotorVelocity();
+        elbowVelocity = getElbowMotorVelocity();
+        time = timer.get();
+        shoulderAcceleration = (shoulderVelocity - prevShoulderVelocity) / Math.abs(time - time2);
+        elbowAcceleration = (elbowVelocity - prevElbowVelocity) / Math.abs(time - time2);
+        prevShoulderVelocity = shoulderVelocity;
+        prevElbowVelocity = elbowVelocity;
+        time2 = timer.get();
+
+        ArmSystemModel.ArmFeedForward tempFeedforward = systemModel.calculateFeedForward(Math.toDegrees(unitModel.toUnits(getShoulderJointAngle())), Math.toDegrees(unitModel.toUnits(getElbowJointAngle())), getShoulderMotorVelocity(), getElbowMotorVelocity(), shoulderAcceleration, elbowAcceleration);
```
katzuv: > `Math.toDegrees(unitModel.toUnits(getShoulderJointAngle())), Math.toDegrees(unitModel.toUnits(getElbowJointAngle()))`

These methods already return values in degrees.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public void periodic() {
+        timer.reset();
+        timer.start();
+        shoulderVelocity = getShoulderMotorVelocity();
+        elbowVelocity = getElbowMotorVelocity();
+        time = timer.get();
+        shoulderAcceleration = (shoulderVelocity - prevShoulderVelocity) / Math.abs(time - time2);
+        elbowAcceleration = (elbowVelocity - prevElbowVelocity) / Math.abs(time - time2);
```
katzuv: Why is `abs()` needed? Is it possible that `time2 < time`?

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,166 @@
...
+        loggerInputs.elbowAngle = getElbowJointAngle();
+        loggerInputs.shoulderAngle = getShoulderJointAngle();
+        loggerInputs.elbowMotorPower = getElbowMotorPower();
+        loggerInputs.shoulderMotorPower = getShoulderMotorPower();
+        loggerInputs.shoulderP = ArmConstants.shoulderP;
+        loggerInputs.shoulderI = ArmConstants.shoulderI;
+        loggerInputs.shoulderD = ArmConstants.shoulderD;
+        loggerInputs.elbowP = ArmConstants.elbowP;
+        loggerInputs.elbowI = ArmConstants.elbowI;
+        loggerInputs.elbowD = ArmConstants.elbowD;
```
katzuv: These are constants, what is the added value in logging them?

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
```diff
@@ -0,0 +1,161 @@
...
+        int n1 = shoulderJointConstants.numberOfMotors;
+        int n2 = elbowJointConstants.numberOfMotors;
+
+        B_MATRIX.set(0, 0, g1 * n1 * FalconConstants.Kt / FalconConstants.R);
+        B_MATRIX.set(1, 1, g2 * n2 * FalconConstants.Kt / FalconConstants.R);
+        B_MATRIX.set(1, 0, 0);
+        B_MATRIX.set(0, 1, 0);
+
+        Kb_MATRIX.set(0, 0, g1 * g1 * n1 * FalconConstants.Kt / FalconConstants.R / FalconConstants.Kv);
+        Kb_MATRIX.set(1, 1, g2 * g2 * n2 * FalconConstants.Kt / FalconConstants.R / FalconConstants.Kv);
```
katzuv, line 58: I'd advise against using the `FalconConstants` here, which I guess originate from the manufacturer's data. Each motor _unit_ is a bit different, and the exact gains should be calculated using SysId.

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
```diff
@@ -0,0 +1,161 @@
...
+        updateCMatrix(theta2, theta1Dot, theta2Dot);
+        updateTgVector(theta1, theta2);
+
+        var thetaDotVector = new SimpleMatrix(2, 1, MatrixType.DDRM);
+        thetaDotVector.set(0, 0, theta1Dot);
+        thetaDotVector.set(1, 0, theta2Dot);
+
+        var thetaDotDotVector = new SimpleMatrix(2, 1, MatrixType.DDRM);
+        thetaDotDotVector.set(0, 0, theta1DotDot);
+        thetaDotDotVector.set(1, 0, theta2DotDot);
```
katzuv, line 130: Perhaps create `thetaDotVector` using the `.copy()` method.

src/main/java/frc/robot/subsystems/arm/commands/JoystickControl.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public JoystickControl(Arm arm, Joystick joystick) {
+        this.arm = arm;
+        this.joystick = joystick;
+    }
+
+    @Override
+    public void execute() {
+        arm.setShoulderJointPower(joystick.getY());
+        arm.setElbowJointPower(joystick.getY());
```
katzuv: Remember, Don't Repeat Yourself. `joystick.getY()` should be stored in a variable.
However, the best way is to have a "throttle" supplier as a parameter of the command, then get its value here. The `XboxControl` command is pretty much identical to this command, a supplier can be used for Xbox input too.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
```diff
@@ -0,0 +1,27 @@
...
+
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.arm.Arm;
+
+public class SetArmsAngles extends CommandBase {
+    private final Arm arm;
+    private final Translation2d position;
+
+    public SetArmsAngles(Arm arm, Translation2d position) {
```
katzuv: This name is confusing. It implies that an angle is a parameter of this class, not a position.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public void execute() {
+        arm.setAngle(position);
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+        arm.setShoulderJointPower(0);
+        arm.setElbowJointPower(0);
```
katzuv: Why is that needed?

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public SetArmsAngles(Arm arm, Translation2d position) {
+        this.arm = arm;
+        this.position = position;
+        addRequirements();
+    }
+
+    @Override
+    public void execute() {
+        arm.setAngle(position);
```
katzuv: This is sufficient to be done in `initialize()`, the Talon "remembers" its setpoint. It's unnecessary to update the setpoint every loop.

src/main/java/frc/robot/subsystems/arm/commands/JoystickControl.java
```diff
@@ -0,0 +1,27 @@
...
+
+import edu.wpi.first.wpilibj.Joystick;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.arm.Arm;
+
+public class JoystickControl extends CommandBase {
+    private final Arm arm;
+    private final Joystick joystick;
+
+    public JoystickControl(Arm arm, Joystick joystick) {
```
katzuv: mussing `addRequirements()`.

src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
```diff
@@ -0,0 +1,26 @@
+package frc.robot.subsystems.arm.commands;
```
katzuv, line 1: Will this and the `SetShoulderAngle` commands be deleted after testing is done?

katzuv (changes requested):
Great job!
There are still some comments left on #7, I recommend just addressing them here.
Also, I'd rename this branch to just `feature/arm`.

src/main/java/frc/robot/Ports.java
Emma03L: I'm pretty sure we only changed the constants to be in the same package as the subsystem, not the ports

src/main/java/frc/robot/subsystems/arm/Arm.java
Emma03L: The configAllSettings() method configures all of the settings to the factory default.

src/main/java/frc/robot/subsystems/arm/Arm.java
Emma03L: Yes I'll delete the ones that aren't used

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
Emma03L: To avoid damaging the motors

src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
Emma03L: Probably

src/main/java/frc/robot/subsystems/arm/Arm.java
Emma03L: In the first loop time2 doesn't has an initial value of 0

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
Emma03L: What's SysId?

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
Emma03L: What would it be copying? There aren't any variables that are identical to thetaDotVector

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,144 @@
...
+    private final ArmKinematics kinematics = new ArmKinematics(ArmConstants.SHOULDER_ARM_LENGTH, ArmConstants.ELBOW_ARM_LENGTH);
+    private final ArmSystemModel systemModel = new ArmSystemModel(ArmConstants.ARM_CONSTANTS);
+
+    private final Timer timer = new Timer();
+
+    private final TalonFX shoulderMainMotor = new TalonFX(Ports.ArmPorts.SHOULDER_MAIN_MOTOR);
+    private final TalonFX shoulderAuxMotor = new TalonFX(Ports.ArmPorts.SHOULDER_AUX_MOTOR);
+    private final TalonFX elbowMainMotor = new TalonFX(Ports.ArmPorts.ELBOW_MAIN_MOTOR);
+    private final TalonFX elbowAuxMotor = new TalonFX(Ports.ArmPorts.ELBOW_AUX_MOTOR);
+    private final TalonSRX shoulderEncoder = new TalonSRX(Ports.ArmPorts.SHOULDER_ENCODER);
```
katzuv: The Talon FXs will just use their internal encoders. You need to configure them to use the Talon SRXs as feedback devices.

katzuv (changes requested):

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,144 @@
...
+    private double shoulderAcceleration;
+    private double elbowAcceleration;
+
+    private Arm() {
+        super(new ArmLogInputs());
+        timer.reset();
+        timer.start();
+
+        shoulderMainMotor.configVoltageCompSaturation(ArmConstants.VOLT_COMP_Saturation);
+        shoulderMainMotor.enableVoltageCompensation(ArmConstants.ENABLE_VOLT_COMPENSATION);
```
katzuv: It makes more sense to first enable the voltage compensation and then configure its level.

src/main/java/frc/robot/subsystems/arm/Arm.java
vichik123: No, it doesn't return the value in degrees. However, it shouldn't be in degrees. If you look at the documentation, it says that all of the parameters are in radians, radians per second, and radians per second squared.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,144 @@
...
+        time2 = timer.get();
+
+        ArmSystemModel.ArmFeedForward tempFeedforward = systemModel.calculateFeedForward(Math.toDegrees(unitModel.toUnits(getShoulderJointAngle())), Math.toDegrees(unitModel.toUnits(getElbowJointAngle())), getShoulderMotorVelocity(), getElbowMotorVelocity(), shoulderAcceleration, elbowAcceleration);
+        shoulderFeedforward = tempFeedforward.shoulderFeedForward;
+        elbowFeedForward = tempFeedforward.elbowFeedForward;
+    }
+
+    public void updateInputs() {
+        loggerInputs.shoulderAngle = getShoulderJointAngle();
+        loggerInputs.elbowAngle = getElbowJointAngle();
```
vichik123, line 213: The calculation should be done in the update inputs method, and the field in the `loggerInputs` should be returned in the functions. This way if the `getShoulderJointAngle()` or `getElbowJointAngle()` methods are called multiple times in the periodic loop, there aren't any duplicate requests to the CanBus.

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
vichik123: We could do SysId

src/main/java/frc/robot/subsystems/arm/commands/XboxControl.java
```diff
@@ -0,0 +1,28 @@
...
+        arm.setShoulderJointPower(xboxController.getLeftY());
+        arm.setElbowJointPower(xboxController.getRightY());
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+        arm.setShoulderJointPower(0);
+        arm.setElbowJointPower(0);
+    }
+}
```
vichik123: This is a default command. There's no reason to add an `end` method as the command never ends.

src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
vichik123: By the way, you're missing the `arm` object in the `addRequirements` method.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -30,9 +36,11 @@ public static RobotContainer getInstance() {
     }
 
     private void configureDefaultCommands() {
+        arm.setDefaultCommand(new XboxControl(arm, xboxController));
     }
 
     private void configureButtonBindings() {
+        a.onTrue(new SetShoulderAngle(arm, 30));
```
vichik123, line 43: This won't work because you have configured a default command for the arm already. What you can do is create a single default command that calls certain operations using ifs. See the `HolonomicDrive` command in the swerve-drive branch.

vichik123 (changes requested):
Overall this looks pretty good. Try to think about how you can switch between joystick control of the arm and the button control. Additionally, figure out how to move the arms up-down and forwards-backwards using the joysticks.

src/main/java/frc/robot/RobotContainer.java
barel13: Why this won't work? Unless it changed this year, when you press a button it overrides the default command, and after finishing it comes back to the default command

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,144 @@
...
+    private double prevShoulderVelocity;
+    private double shoulderFeedforward;
+    private double prevElbowVelocity;
+    private double elbowFeedForward;
+    private double time;
+    private double time2 = 0;
+    private double shoulderVelocity;
+    private double elbowVelocity;
+    private double shoulderAcceleration;
+    private double elbowAcceleration;
```
barel13: As you only use this for calculating the acceleration, it might be worth creating a simple class that calculates the acceleration and updates automatically when new velocity is sampled

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,144 @@
...
+    public double getShoulderMotorVelocity() {
+        return unitModel.toVelocity(shoulderMainMotor.getSelectedSensorVelocity());
+    }
+
+    public double getElbowMotorVelocity() {
+        return unitModel.toVelocity(elbowMainMotor.getSelectedSensorVelocity());
+    }
+
+    public String getSubsystemName() {
+        return "PrototypeArm";
```
barel13: It's not a prototype anymore🥳

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final boolean clockWise = true;
+
+    public static final double shoulderP = 0.1;
+    public static final double shoulderI = 0.0;
+    public static final double shoulderD = 0.0;
+    public static final double elbowP = 0.1;
+    public static final double elbowI = 0.0;
+    public static final double elbowD = 0.0;
+
+    public static final double DEAD_BEND = 0.05;
```
barel13: I'm sorry `DEADBAND`

src/main/java/frc/robot/subsystems/arm/commands/JoystickControl.java
barel13: Also, the writers of the controllers are idiots and idk why they chose that down is positive so you should invert the value

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,27 @@
...
+import frc.robot.subsystems.arm.Arm;
+
+public class SetArmsPosition extends CommandBase {
+    private final Arm arm;
+    private final Translation2d position;
+
+    public SetArmsPosition(Arm arm, Translation2d position) {
+        this.arm = arm;
+        this.position = position;
+        addRequirements();
```
barel13: Require the arm

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        arm.setPosition(position);
+    }
+
```
barel13, line 23: You should finish when you reach the angle

src/main/java/frc/robot/subsystems/arm/commands/XboxControl.java
```diff
@@ -0,0 +1,28 @@
...
+    public XboxControl(Arm arm, XboxController xboxController) {
+        this.arm = arm;
+        this.xboxController = xboxController;
+        addRequirements(arm);
+    }
+
+    @Override
+    public void execute() {
+        arm.setShoulderJointPower(xboxController.getLeftY());
+        arm.setElbowJointPower(xboxController.getRightY());
```
barel13: Same about the inversion

barel13 (changes requested):

src/main/java/frc/robot/subsystems/arm/commands/XboxControl.java
katzuv: Default commands end when they are interrupted by another command.

src/main/java/frc/robot/subsystems/arm/ArmLogInputs.java
```diff
@@ -0,0 +1,28 @@
...
+        table.put("elbowAngle", elbowAngle);
+        table.put("shoulderAngle", shoulderAngle);
+        table.put("elbowMotorPower", elbowMotorPower);
+        table.put("shoulderMotorPower", shoulderMotorPower);
+    }
+
+    public void fromLog(LogTable table) {
+        elbowAngle = table.getDouble("elbowAngle", elbowAngle);
+        shoulderAngle = table.getDouble("shoulderAngle", shoulderAngle);
+        elbowMotorPower = table.getDouble("elbowMotorPower", elbowMotorPower);
```
GaiaZano05: You should had angle setPoint for both arms.

src/main/java/frc/robot/subsystems/arm/Arm.java
Emma03L: Do you mean the acceleration calculation that is currently in the periodic?

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
Emma03L: There is an end method, once the arm reaches the position it will stop

barel13: Yeah, but because you use commands, you should override the `isFinished()` method

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,132 @@
...
+        super(new ArmLogInputs());
+        shoulderMainMotor.configVoltageCompSaturation(ArmConstants.VOLT_COMP_Saturation);
+        shoulderMainMotor.enableVoltageCompensation(ArmConstants.ENABLE_VOLT_COMPENSATION);
+        shoulderMainMotor.setNeutralMode(NeutralMode.Brake);
+        shoulderMainMotor.setInverted(ArmConstants.clockWise);
+        shoulderMainMotor.config_kP(0, ArmConstants.shoulderP);
+        shoulderMainMotor.config_kI(0, ArmConstants.shoulderI);
+        shoulderMainMotor.config_kD(0, ArmConstants.shoulderD);
+        shoulderMainMotor.configNeutralDeadband(ArmConstants.DEADBAND);
+//        shoulderMainMotor.configSelectedFeedbackSensor(RemoteFeedbackDevice.valueOf(shoulderEncoder.));
```
katzuv: Don't commit commented-out code. Commit it (with an indicative commit message) and delete it. If you'll need it in the future. just `git revert` that commit.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,132 @@
...
+public class Arm extends LoggedSubsystem<ArmLogInputs> {
+    private static Arm INSTANCE = null;
+
+    private final ArmKinematics kinematics = new ArmKinematics(ArmConstants.SHOULDER_ARM_LENGTH, ArmConstants.ELBOW_ARM_LENGTH);
+    private final ArmSystemModel systemModel = new ArmSystemModel(ArmConstants.ARM_CONSTANTS);
+    private final ArmAccelerationCalculation accelerationCalculation = new ArmAccelerationCalculation();
+
+    private double shoulderFeedforward;
+    private double elbowFeedForward;
+    private double[] accelerations;
```
katzuv: If you have separate variables for the parts of the arm, I'd store the accelerations in different variables too.

src/main/java/frc/robot/subsystems/LoggedSubsystem.java
```diff
@@ -28,4 +28,4 @@ public void updateSubsystem() {
     public abstract void updateInputs();
 
     public abstract String getSubsystemName();
-}
+}
```
katzuv, line 31: Why is this change made in this branch?

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,132 @@
...
+    public void setElbowJointPower(double power) {
+        elbowMainMotor.set(TalonFXControlMode.PercentOutput, power);
+    }
+
+    public void setShoulderJointAngle(double angle) {
+        shoulderMainMotor.set(TalonFXControlMode.MotionMagic, unitModel.toTicks(Math.toRadians(angle)), DemandType.ArbitraryFeedForward, shoulderFeedforward);
+        shoulderSetPoint = angle;
+    }
+
+    public void setElbowJointAngle(double angle) {
```
katzuv: Use `Rotation2d` as the parameters' types.

src/main/java/frc/robot/subsystems/arm/Arm.java
katzuv: > The calculation should be done in the update inputs method

I don't think that's a good idea. That method should only update inputs, not do any other calculations. These should be done, perhaps, in the `periodic()` method.
You could save the calculation result in a class field if you want to save a few status frames (the messages sent over the CAN bus).

src/main/java/frc/robot/subsystems/arm/Arm.java
katzuv: > No, it doesn't return the value in degrees

In the time I reviewed the code, they did:
https://github.com/Galaxia5987/robot-2023/blob/183910557936f7553cdf35ec6c11c034f1f82ae6/src/main/java/frc/robot/subsystems/arm/Arm.java#L108-L114
I still don't see that documentation.
Anyway, I'm not sure why not work with degrees. Looks more convenient to me, user-wise.

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
katzuv: > What's SysId?

https://docs.wpilib.org/en/stable/docs/software/pathplanning/system-identification/introduction.html

src/main/java/frc/robot/subsystems/arm/commands/JoystickControl.java
katzuv: So maybe if you don't know the reason, it's a bit too early to call them idiots?

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
katzuv: Whoops. I missed that you use `theta1DotDot` instead of `theta1Dot` in the second matrix.
Otherwise, it'd make sense to copy the matrix, because they would we identical.

src/main/java/frc/robot/subsystems/arm/Arm.java
katzuv: In the latest version of the code (2e30503) it does, but I still don't understand how an absolute value helped to remedy that.
Anyway, I had a few notes about all the time stuff [here](https://github.com/Galaxia5987/robot-2023/pull/7#discussion_r1083065790).

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
katzuv: Why? Will the stall torque damage them?
Anyway, I don't think that just uncontrollably "shutting down" the arm is a good idea.
This also just raised an important question -- what is the "neutral" position of the arm, i.e., when the system has no power? For example, before a match starts and after it ends,

src/main/java/frc/robot/subsystems/arm/ArmAccelerationCalculation.java
```diff
@@ -0,0 +1,37 @@
+package frc.robot.subsystems.arm;
+
+import edu.wpi.first.wpilibj.Timer;
+import frc.robot.utils.units.UnitModel;
+
+public class ArmAccelerationCalculation {
```
katzuv: This class and the method it contains can be static.

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Forward kinematics for the arm.
+     *
+     * @param shoulderAngle the angle of the shoulder joint. [rad]
+     * @param elbowAngle the angle of the elbow joint. [rad]
+     * @return the position of the end effector. ([m, [m])
+     */
+    public Translation2d forwardKinematics(double shoulderAngle, double elbowAngle) {
```
katzuv, line 27: IMO using `Rotation2d`s here is better.

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Inverse kinematics for the arm.
+     *
+     * @param x the x coordinate of the end effector. [m]
+     * @param y the y coordinate of the end effector. [m]
+     * @return the angles of the shoulder and elbow joints. ([rad, rad])
+     */
+    public InverseKinematicsSolution inverseKinematics(double x, double y) {
```
katzuv: IMO using `Translation2d` here is better.

katzuv (changes requested):

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
Emma03L: When I think about it the end method shouldn't really matter because it will just return to the default command (xbox/joystick control)

src/main/java/frc/robot/subsystems/arm/Arm.java
Emma03L: Why

src/main/java/frc/robot/subsystems/arm/ArmKinematics.java
Emma03L: why?

katzuv: https://github.com/Galaxia5987/robot-2023/pull/7#discussion_r1082344997

src/main/java/frc/robot/subsystems/arm/Arm.java
katzuv: https://github.com/Galaxia5987/robot-2023/pull/7#discussion_r1082344997

src/main/java/frc/robot/subsystems/arm/Arm.java
katzuv: You're probably talking about the `configFactoryDefault()` method.

src/main/java/frc/robot/subsystems/arm/commands/SetArmsAngles.java
katzuv: When running the Joystick or Xbox control commands, if the controller isn't touched, the arm will drop (the code will essentially be `motor.set(TalonFXControlMode.PercentOutput, 0)`). Is this the desired behaviour?

katzuv: Also, dropping a link here to my comment about reducing the Joystick or Xbox "manual" control commands into one command: https://github.com/Galaxia5987/robot-2023/pull/12#discussion_r1084659696.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+        return INSTANCE;
+    }
+
+    /**
+     * Configures the aux motors
+     * @param shoulderAuxMotor
+     * @param shoulderMainMotor
+     */
+    private void configureAuxMotor(TalonFX shoulderAuxMotor, TalonFX shoulderMainMotor) {
```
barel13: I see we forgot to rename the parameters😂

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public void initialize() {
+        arm.setEndPosition(position);
+    }
+
+    @Override
+    public boolean isFinished() {
+       return MathUtil.applyDeadband(position.getNorm(), ArmConstants.SETPOINT_DEADBAND) == arm.getEndPosition().getNorm();
```
barel13: The norm is the magnitude of the position, which is not what you want to check, I think it will be easier to just check k the x component and the y component

src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        arm.setElbowJointAngle(angle);
+    }
+
+    @Override
+    public boolean isFinished() {
+        return MathUtil.applyDeadband(angle, ArmConstants.SETPOINT_DEADBAND) == arm.getEndPosition().getNorm();
```
barel13: I guess that you didn't have time to change it yesterday, but just in case, you should check against the angle of the elbow

barel13 (changes requested):

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
Emma03L: So should I use the applyDeadband twice? once for the x component and once for the y

barel13: Unfortunately yes

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,217 @@
...
+        return INSTANCE;
+    }
+
+    /**
+     * Configures the aux motors
+     * @param auxMotor
+     * @param mainMotor
+     */
+    private void configureAuxMotor(TalonFX auxMotor, TalonFX mainMotor) {
+        auxMotor.follow(shoulderMainMotor);
```
barel13: It should follow the main motor

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        arm.setEndPosition(position);
+    }
+
+    @Override
+    public boolean isFinished() {
+       return MathUtil.applyDeadband(position.getX(), ArmConstants.SETPOINT_DEADBAND) == arm.getEndPosition().getX() && MathUtil.applyDeadband(position.getY(), ArmConstants.SETPOINT_DEADBAND) == arm.getEndPosition().getY();
```
barel13: I'm really sorry for misleading you yesterday, the way you check that they have reached the setpoint is by checking that the difference is in the deadband `applyDeadband(Ax-Bx) == 0` (pseudocode). It applies to the other commands as well

barel13 (changes requested):
Great job, after you finish this review I'll approve it🥳

src/main/java/frc/robot/subsystems/arm/commands/ArmJoystickControl.java
```diff
@@ -0,0 +1,27 @@
...
+    public ArmJoystickControl(Joystick joystick) {
+        this.arm = Arm.getInstance();
+        this.joystick = joystick;
+        addRequirements(arm);
+    }
+
+    @Override
+    public void execute() {
+        arm.setShoulderJointPower(-joystick.getY());
+        arm.setElbowJointPower(-joystick.getY());
```
barel13: The same joystick will move both joints?

barel13 (approved):

build.gradle
```diff
@@ -92,6 +92,7 @@ dependencies {
     implementation 'com.sparkjava:spark-template-velocity:2.7.1'
 
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.9'
+    annotationProcessor "org.littletonrobotics.akit.junction:junction-autolog:2.1.0"
```
katzuv, line 95: This should be added in another branch, as it's relevant to all the code and not only to the arm.

src/main/java/frc/robot/subsystems/arm/AccelerationCalculation.java
```diff
@@ -0,0 +1,21 @@
+package frc.robot.subsystems.arm;
+
+
+public class AccelerationCalculation {
+    private double currentVelocity;
+    private double prevVelocity;
+    private double currentTime;
+    private double prevTime;
+
+    public void addVelocity(double sVelocity, double time) {
```
katzuv, line 10: What does "s" stand for?

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,219 @@
...
+     * @param mainMotor
+     */
+    private void configureAuxMotor(TalonFX auxMotor, TalonFX mainMotor) {
+        auxMotor.follow(mainMotor);
+        auxMotor.enableVoltageCompensation(ArmConstants.ENABLE_VOLT_COMPENSATION);
+        auxMotor.configVoltageCompSaturation(ArmConstants.VOLT_COMP_SATURATION);
+        auxMotor.configMotionAcceleration(ArmConstants.MOTION_ACCELERATION);
+        auxMotor.configMotionCruiseVelocity(ArmConstants.MOTION_CRUISE_VELOCITY);
+        auxMotor.setNeutralMode(NeutralMode.Brake);
+    }
```
katzuv, line 68: Use the `TalonFXConfiguration` class instead.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,219 @@
...
+        mainMotor.configVoltageCompSaturation(ArmConstants.VOLT_COMP_SATURATION);
+        mainMotor.setNeutralMode(NeutralMode.Brake);
+        mainMotor.setInverted(ArmConstants.CLOCKWISE);
+
+        mainMotor.config_kP(0, kP);
+        mainMotor.config_kI(0, kI);
+        mainMotor.config_kD(0, kD);
+
+        mainMotor.configNeutralDeadband(ArmConstants.DEADBAND);
+        mainMotor.setSelectedSensorPosition(unitModel.toTicks(encoder.getAbsolutePosition()));
```
katzuv, line 89: This line runs once for each motor. You want the Talon FX to continuously read measurements from the CANCoder.

src/main/java/frc/robot/subsystems/arm/AccelerationCalculation.java
```diff
@@ -0,0 +1,21 @@
+package frc.robot.subsystems.arm;
+
+
+public class AccelerationCalculation {
+    private double currentVelocity;
+    private double prevVelocity;
+    private double currentTime;
+    private double prevTime;
```
katzuv, line 8: Don;t shorten "previous", use the full form of the word.

src/main/java/frc/robot/subsystems/arm/AccelerationCalculation.java
```diff
@@ -0,0 +1,21 @@
+package frc.robot.subsystems.arm;
+
+
+public class AccelerationCalculation {
```
katzuv, line 4: `AccelerationCalculator` fits better IMO.

src/main/java/frc/robot/subsystems/arm/Arm.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private final TalonFX elbowMainMotor = new TalonFX(Ports.ArmPorts.ELBOW_MAIN_MOTOR);
+    private final TalonFX elbowAuxMotor = new TalonFX(Ports.ArmPorts.ELBOW_AUX_MOTOR);
+    private final CANCoder elbowEncoder = new CANCoder(Ports.ArmPorts.ELBOW_ENCODER);
+
+    private final UnitModel unitModel = new UnitModel(ArmConstants.TICKS_PER_RADIAN);
+    private double shoulderFeedforward;
+    private double elbowFeedForward;
+    private double shoulderSetpoint;
+    private double elbowSetpoint;
```
katzuv, line 34: I'd zero all of those values.

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,47 @@
+package frc.robot.subsystems.arm;
+
+import com.ctre.phoenix.motorcontrol.TalonFXInvertType;
+
+public class ArmConstants {
```
katzuv, line 5: `Constants` is enough, this class is already inside the arm pacakge.

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,47 @@
...
+
+    // Arm constants
+    public static final double VOLT_COMP_SATURATION = 10; //[V]
+    public static final boolean ENABLE_VOLT_COMPENSATION = true;
+    public static final double MOTION_ACCELERATION = 0;
+    public static final double MOTION_CRUISE_VELOCITY = 0;
```
katzuv, line 11: Are these the same for both parts of the arm?

src/main/java/frc/robot/subsystems/arm/ArmConstants.java
```diff
@@ -0,0 +1,47 @@
+package frc.robot.subsystems.arm;
+
+import com.ctre.phoenix.motorcontrol.TalonFXInvertType;
+
+public class ArmConstants {
+
+    // Arm constants
+    public static final double VOLT_COMP_SATURATION = 10; //[V]
+    public static final boolean ENABLE_VOLT_COMPENSATION = true;
```
katzuv, line 9: 1. It makes more sense to first enable the voltage compensation, then set the specific value.
2. Don't shorten "voltage", use the full form.

src/main/java/frc/robot/subsystems/arm/ArmInputs.java
```diff
@@ -0,0 +1,15 @@
+package frc.robot.subsystems.arm;
+
+import org.littletonrobotics.junction.AutoLog;
+import org.littletonrobotics.junction.LogTable;
+import org.littletonrobotics.junction.inputs.LoggableInputs;
+
+@AutoLog
```
katzuv, line 7: Great!

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
katzuv: @barel13, my intention was to let @Emma03L read my comment, then she could resolve the conversation after she would have read it.

src/main/java/frc/robot/subsystems/arm/ArmSystemModel.java
```diff
@@ -0,0 +1,161 @@
...
+    private final SimpleMatrix B_MATRIX = new SimpleMatrix(2, 2, MatrixType.DDRM);
+
+    /**
+     * The constructor for the system model.
+     *
+     * @param systemConstants the system constants of the arm.
+     */
+    public ArmSystemModel(SystemConstants systemConstants) {
+        var shoulderJointConstants = systemConstants.getShoulderJointConstants();
+        var elbowJointConstants = systemConstants.getElbowJointConstants();
```
katzuv, line 39: This is the only place the `SystemConstants` class is used, and you just unpack it here (so you don't really use the joined (😉) class.
I'd delete it and use the `JointConstants` objects directly.

src/main/java/frc/robot/subsystems/arm/SystemConstants.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public JointConstants getElbowJointConstants() {
+        return elbowJointConstants;
+    }
+
+    public static class JointConstants {
+        public double mass;
+        public double length;
+        public double momentOfInertia;
+        public double cmRadius;
```
katzuv, line 30: `radiusCm` sounds better, but if you don't include the units in symbols (functions, variables, etc.) anywhere else, I'd omit it here too.

src/main/java/frc/robot/subsystems/arm/SystemConstants.java
```diff
@@ -0,0 +1,43 @@
...
+        public double gearing;
+        public int numberOfMotors;
```
katzuv, line 32: `motorsNumber` is shorter and just as readable.

src/main/java/frc/robot/subsystems/arm/commands/ArmJoystickControl.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.arm.Arm;
+
+public class ArmJoystickControl extends CommandBase {
+    private final Arm arm;
+    private final Joystick shoulderJoystick;
+    private final Joystick elbowJoystick;
+
+    public ArmJoystickControl(Joystick shoulderJoystick, Joystick elbowJoystick) {
+        this.arm = Arm.getInstance();
```
katzuv, line 13: Pass the `Arm` object as a parameter to this command (as it's done in all other commands).

src/main/java/frc/robot/subsystems/arm/commands/ArmJoystickControl.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.wpi.first.wpilibj.Joystick;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.arm.Arm;
+
+public class ArmJoystickControl extends CommandBase {
+    private final Arm arm;
+    private final Joystick shoulderJoystick;
+    private final Joystick elbowJoystick;
+
+    public ArmJoystickControl(Joystick shoulderJoystick, Joystick elbowJoystick) {
```
katzuv, line 12: It's better to pass suppliers of the throttle value instead of the joysticks themselves.

src/main/java/frc/robot/subsystems/arm/commands/ArmJoystickControl.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public void execute() {
+        arm.setShoulderJointPower(-shoulderJoystick.getY());
+        arm.setElbowJointPower(-elbowJoystick.getY());
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+        arm.stop();
```
katzuv, line 27: This will make the arm drop in a non-graceful way. Is this the desired behavior?

src/main/java/frc/robot/subsystems/arm/commands/SetArmsPosition.java
```diff
@@ -0,0 +1,28 @@
+package frc.robot.subsystems.arm.commands;
+
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.arm.Arm;
+import         edu.wpi.first.math.MathUtil;
```
katzuv, line 6: Running <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>L</kbd> will fix the issue.

src/main/java/frc/robot/subsystems/arm/commands/SetElbowAngle.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        arm.setElbowJointAngle(angle);
+    }
+
+    @Override
+    public boolean isFinished() {
+        return MathUtil.applyDeadband(angle - arm.getElbowJointAngle(), ArmConstants.SETPOINT_DEADBAND) == 0;
```
katzuv, line 25: 1. This is not the same as deadband, "tolerance" is a better name (talking about the constant).
2. You don't need `MathUtil.applyDeadband()` here, you can just check whether `abs(angle - arm.getElbowJointAngle()) < Constants.SETPOINT_TOLERANCE`.
3. You repeat this logic between commands. Extract it into a method in the `Arm` class.
4. You should check for the velocity, too, and not just for the position. If the elbow passes through its setpoint but at a high speed, it will overshoot.

src/main/java/frc/robot/subsystems/arm/commands/SetShoulderAngle.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Override
+    public boolean isFinished() {
+        return MathUtil.applyDeadband(angle - arm.getShoulderJointAngle(), ArmConstants.SETPOINT_DEADBAND) == 0;
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+        arm.stop();
```
katzuv, line 30: Do you want to stop the whole arm when this commands control the shoulder only?

src/main/java/frc/robot/utils/motors/FalconConstants.java
```diff
@@ -0,0 +1,12 @@
+package frc.robot.utils.motors;
+
+public class FalconConstants {
```
katzuv, line 3: Use WPILib's `DCMotor` class instead.

katzuv (changes requested):
Getting there!

Emma03L:
merged branch

PR #17, opened 2023-01-31
Pick command group

AlonSchwierz:


src/main/java/frc/robot/subsystems/intake/Command/Feed.java
```diff
@@ -1,15 +1,18 @@
...
 
 public class Feed extends CommandBase {
     private final Intake intake;
+    private final BeamBreaker beamBreaker;
     private final double power;
     private final double angle;
 
     public Feed(double power, double angle) {
         this.intake = Intake.getInstance();
+        this.beamBreaker = BeamBreaker.getInstance();
```
katzuv: Subsystems should be passed as parameters to commands.

src/main/java/frc/robot/commandGroup/PickUpCube.java
```diff
@@ -0,0 +1,22 @@
...
+
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.subsystems.arm.commands.SetArmsPosition;
+import frc.robot.subsystems.gripper.Gripper;
+import frc.robot.subsystems.intake.Command.Feed;
+
+public class PickUpCube extends SequentialCommandGroup {
+    public PickUpCube(Gripper gripper, double intakePower, Translation2d armPosition) {
```
katzuv: Why do the arm position and intake power need to be passed as a parameter?
On the other hand, the arm should be passed as a parameter.

src/main/java/frc/robot/commandGroup/PickUpCube.java
```diff
@@ -0,0 +1,22 @@
...
+        addCommands(
+                // Feeds the cube into the Intake until the BeamBreaker sends a false signal
+                new Feed(intakePower, 0),
```
katzuv: Testing in the future may show that an angle other than 0 should be used. The angle should be passed as a parameter or saved as a constant.

katzuv (changes requested):

katzuv:
Stuff here (branch, PR title, command) should be renamed to "pick up from floor".
Also, the branch name should be prefixed by `feature` and the words "command group" are redundant – the branch and PR names should explain _what_ the code does, not _how_ it does it.

src/main/java/frc/robot/subsystems/intake/Command/Feed.java
AlonSchwierz: done.

src/main/java/frc/robot/commandGroup/PickUpCube.java
AlonSchwierz: done.

src/main/java/frc/robot/subsystems/intake/Command/Feed.java
katzuv: What about the arm?
Also, don't resolve conversations by yourself – it's the reviewer's role, so they can see the fix actually fixed the problem (which didn't fully happen here).

src/main/java/frc/robot/commandGroup/PickUpCube.java
katzuv: Why did you resolve this conversation?

PR #27, opened 2023-03-29
Simpler led functions

vichik123:
This pull requests adds cool new LED functions that make them blink-a-blink-blink


pklito:
Just came to say I love the term "blink-a-blink-blink" 👍

src/main/java/frc/robot/subsystems/leds/Leds.java
```diff
@@ -67,31 +95,44 @@ public boolean inConeMode() {
...
+            for (int i = 0; i < ledBuffer.getLength(); i++) {
+                ledBuffer.setHSV(i, rainbowHue, 255, 180);
+                rainbowHue += (180 / ledBuffer.getLength());
+                leds.setData(ledBuffer);
+                System.out.println("in-loop check statement");
+            rainbowHue%=180;
+            }
+            System.out.println("after loop statement");
+        }
     }
```
pklito, line 133: Take this with a grain of salt since it's been forever since ive worked in FRC,
but shouldn't if rainbow be checked _before_ the Primary and Secondary colors?

In this scenario for 1 frame you will have a flickering of the leds.

### Suggestion: (i forgot how to actually write it in the code:
above `if(blink)` (line 96 or after in line 105)
`if(isRainbow) {mode = Mode.ELSE}`
then you just have a simple
```java
if(PRIMARY) {}
else if (SECONDARY) {}
else{/*ELSE*/}
```

Emma03L: The rainbow LEDs work great so I guess if it ain't broke don't fix it

katzuv: This is a good rule to follow many times, but not always. The offseason is the best time to improve software and learn, now's the time to make such changes and improvements to the code.

katzuv: This is a good rule to follow many times, but not always. The offseason is the best time to improve software and learn, now's the time to make such changes and improvements to the code.

